PR #8, opened 2016-04-26
Dev cleanup

justinvdk:
removed all loose src dirs. All code so far now lives in the PL2 maven project.


wouterjsmit:
Obviously an improvement from the unmanaged code. Commit Titles/messages bit messy but a single version is more important now.

Merging.

PR #9, opened 2016-04-27
Added the test annotation classes

FarisElghlan:
Added the test annotation classes, which I've shown last week during a meeting, to the project. Check the AccessPrivateTest and PrivateMemberClass test classes in the test package of the shared module to see how to use the test annotation.

### DIFF
--- a/PL2/PL2-gui/pom.xml
+++ b/PL2/PL2-gui/pom.xml
@@ -17,6 +17,12 @@
             <artifactId>PL2-shared</artifactId>
             <version>${PL2-shared-version}</version>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.10</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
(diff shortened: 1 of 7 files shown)


wouterjsmit:
Great! Obviously I have no comments on your code. We might want to still consider packaging it and importing via Maven, as we discussed earlier this week. Did you find any way to do this, or did you conclude that it is too much work?

Only comment is the `pom` changes. @justinvdk probably knows best how the maven modules inherit, but I recall that each module inherits from the parent `pom`, so we might want to put the `JUnit` dependency in there instead. What's your advice on that @justinvdk? If Justin agrees with the `pom` setup, and location of these files in the project, this seems ready for merging.

P.S. the diff shows that the `pom`s do not end with a newline. They note this as a style error, so I'm guessing there should be one? Not entirely sure what the standard for this is.

FarisElghlan:
I did look at how to create a maven dependency, but it seems like you have to upload your code to a central maven repository and are not allowed to change it after that (so you would have to create a new version if anything changes). This seems like a bit too much effort to me and I don't have much experience with maven, so I didn't completely understand everything which had to be done to host a dependency.

justinvdk:
@FarisElghlan is right. Maven dependencies in the test scope are not transitive. So for junit we indeed need to declare the dependency on a per-module basis. This also goes for sources, and as we (indeed) do not manage our own (deployed) maven dependency the most convenient solution is to just place them into the main sources of -shared.

Merge fine by me

PR #11, opened 2016-04-28
Add all files in .idea/ to gitignore

wouterjsmit:
We do not want to share project files whatsoever.


wouterjsmit:
Closing because addition not sufficient.

PR #19, opened 2016-04-29
Temp graph visualizer

FarisElghlan:
Added visualization of graphs and added the jars to the deliverables.


wouterjsmit:
All righto, happy release

PR #20, opened 2016-04-29
Bug fix: tree could get GCed.

FarisElghlan:
The tree object could be garbage collected by the JVM,  which called the finalize method on the tree class (which is an external dependency), which then cleared the content of some of the child array lists. This is fixed by keeping a reference to the tree object in the main class.


cathmer:
If this fixes the bug with the two screens I think it can be merged.

PR #25, opened 2016-05-03
Add 10 genome tree file

justinvdk:
In this file only the genomes that are found in our gfa 'dummy' file


wouterjsmit:
Great, merging

PR #27, opened 2016-05-03
bend lines a bit

justinvdk:
As actually avoiding collisions in non-trivial, there is no realistic way to do that for our graph. This change does bend the curved a bit by a fixed amount to better distinguish between two different lines (now curves).

### DIFF
--- a/PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/tempGraph/DrawGraph.java
+++ b/PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/tempGraph/DrawGraph.java
@@ -5,15 +5,21 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
+
+import javafx.beans.value.ChangeListener;
+import javafx.geometry.Point2D;
 import javafx.scene.Scene;
 import javafx.scene.control.Label;
 import javafx.scene.control.ScrollPane;
 import javafx.scene.layout.Pane;
+import javafx.scene.paint.Color;
 import javafx.scene.shape.Circle;
-import javafx.scene.shape.Line;
+import javafx.scene.shape.CubicCurve;
+import javafx.scene.shape.QuadCurve;
 import javafx.stage.Stage;
 import static nl.tudelft.pl2016gr2.core.algorithms.AlgoRunner.FILENAME;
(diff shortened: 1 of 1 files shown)


justinvdk:
As discussed, no real trivial method to prevent lines colliding. All methods in laying out graphs in a visually pleasing way revolve around positioning the nodes properly

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/tempGraph/DrawGraph.java
```diff
@@ -113,18 +119,68 @@ private void drawEdges(Pane pane, HashMap<Integer, Bubble> bubbles, HashMap<Inte
...
+        }
+
+        curve.setControlX(perpP.getX());
+        curve.setControlY(perpP.getY());
+    }
+
+    /**
+     *
+     * Not that 4(!) objects are created for a single invocation. This is considerably slower than
+     * a more efficient way of handling thing. (atleast until the JIT kicks in it seems)
```
wouterjsmit, line 167: Obviously anything is slower than a more efficient way of doing the same thing. :+1:

cathmer:
I guess this is okay for now, as long as we haven't finalized the positioning of the nodes and the zooming it probably doesn't make sense to put more logic in to this. In a later stage this might be necessary though, because I do think it's possible to put some logic in the edges to prevent more collissions.

wouterjsmit:
Most likely the bubble algorithm will reduce the collision problem. If we can manage to have a limited amount of bubbles active on the screen, there will be less edges and collision automatically is less of a problem.

We should also consider completely hiding edges that span the whole screen (a large inversion for instance). These are not relevant when zoomed in anyway, and need only be shown in a high level view of the graph.

JavaFX not my thing, but the code seems fine. Maybe avoid the instantiation of 4 objects by doing that calculation manually?

FarisElghlan:
It is a nice starting point, but currently the curves don't avoid collision and EVERY edge is curved (even very short ones with no collisions), so this definetly needs a lot of adjustment to meet the performance criteria.
I am not sure if we should merge this branch into dev. The code can be used as a reference when implmenting curved edges in our final product, but at the moment I feel like we shouldn't use this code yet, as too many other things have to be done before we should worry about curved edges (semantic zooming) and in the current state the curved edges only cause a penalty in performance and don't make the graph easier to read.

Instantiating 4 objects really shouldn't be a problem at all. If there are any performance issues because of curved edge calculations it will be because the calculations are done too often and not because the calculations take like 0.1% more time because of object creation.

PR #32, opened 2016-05-04
Add feedback

jimhommes:
Feedback can be found in the attached file.


wouterjsmit:
Merging when squashed, thanks for the feedback.

PR #33, opened 2016-05-04
Slightly changed checkstyles naming convention.

FarisElghlan:
I changed the checkstyle.xml file so it allows names starting with 1 lower case character, directly followed by an upper case character (i.e. "xPos"). First it only allowed names which start with at least 2 lower case characters.
I also banned the use of the numbers 0-9 in method/field/parameter/local variable names and banned the use of the '_' character in method names, as this is in my opinion not nice to do in java.

### DIFF
--- a/PL2/checkstyle.xml
+++ b/PL2/checkstyle.xml
@@ -94,7 +94,7 @@
             <property name="option" value="EOL"/>
         </module>
         <module name="PackageName">
-            <property name="format" value="^[a-z]+(\.[a-z][a-z]*)*$"/>
+            <property name="format" value="^[a-z]+(\.[a-z][a-z0-9]*)*$"/>
             <message key="name.invalidPattern"
              value="Package name ''{0}'' must match pattern ''{1}''."/>
         </module>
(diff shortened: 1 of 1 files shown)


wouterjsmit:
Great. Could you squash the commits into one and give that one commit a descriptive message?
Title should be something like "allow more Capitals and prohibit numbers" and a message (wrap every 72 characters) containing the exact changes to the rules you made, in a bullet list if you so prefer:
- allow a single lowercase character to be followed by an uppercase character
- Prohibit numbers in methodnames
- etc

Commits like these (on the convention rules and other 'setup' files) are clear now, but in retrospective will become vague and not understood fast, unless very explicitly documented.

FarisElghlan:
I have merged this branch multiple times into a branch I was working on, because I was busy fixing all checkstyle and PMD warnings (while fixing checkstyle errors I came to the conclusion that I had to change some more checkstyle settings), so I don't think it is a good idea to squash the commits.
Currently the commits/mergen look like this:
commit history
As you can see, after every commit I merged this branch into my other branch (where I was working on fixing chechstyle errors), so I assume squashing the commits will cause problems (and is not worth the risk/extra effort imo).

wouterjsmit:
I see, this will indeed cause a lot of trouble for your other branch.

In the future, I would advice to not merge side branches into each other as much, since if this checkstyle change proposal was rejected, you would also have to amend your branch.

Another fact I just noticed: other/ is not a naming convention, misc/ is what you want.

In any way, merging as there is no real need to prolong this any further.

Issue #34, opened 2016-05-04
Checkstyle adjustments


FarisElghlan:
Currently got 2 problems with checkstyle:
1. Checkstyle currently doesn't seem to check for maximum method length.
2. Checkstyle complains about multiple variable declarations on the same line, for example:
    double d1, d2;
    is wrong and must be changed to:
    double d1;
    double d2;
    In my opinion the first option is better, because it is shorter and still easy to read.

wouterjsmit:
[The SUN Java convention](http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141270.html#2992) is to use one declaration per line.

However, I personally am a fan of multiple declarations, as I think variables should be self-documenting (which is encouraged with the rather strict naming conventions for variables by PMD and Checkstyle).

I opened a a PR #36 with the integration of the MaximumMethodLength module.

If others agree, I will also remove the MultipleVariableDeclarations modules (to allow multiple variables on one line).

FarisElghlan:
As stated in my first post I agree with removing the MultipleVariableDeclarations rule, but I guess that comment was intended for the other people in the group. Maybe tell them via whatsapp to post their opinion on this rule here?

cbilstra:
I agree that multiple variable declarations should be possible.

justinvdk:
For Java I too think it should be possible.
I think devs are anxious because in C it can get tricky with multi variable declaration in combination with pointers.

PR #36, opened 2016-05-05
Amend Checkstyle

wouterjsmit:
Fixes: #34

This is a common check that missed in the default 'google_checks'.
The maximum length is set to 30. (more or less arbitrarily)

### DIFF
--- a/PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/FilterSnipsTest.java
+++ b/PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/FilterSnipsTest.java
@@ -48,7 +48,6 @@ public class FilterSnipsTest {
       } else {
         node.addInlink(i - 2);
       }
-
       graph.addNode(node);
     }
   }
--- a/PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawGraph.java
+++ b/PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawGraph.java
@@ -14,7 +14,6 @@ import nl.tudelft.pl2016gr2.parser.controller.FullGfaReader;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.Set;
 
 /**
@@ -165,20 +164,7 @@ public class DrawGraph {
     nodes.forEach((Integer id, Node node) -> {
       nodeDepths.put(id, new NodeDepth(node));
     });
-    Set currentNode = new HashSet<>();
-    currentNode.add(root.getId());
-    int depth = 0;
-    while (!currentNode.isEmpty()) {
-      Set nextNodes = new HashSet();
-      Iterator<Integer> it = currentNode.iterator();
-      while (it.hasNext()) {
-        int currentNodes = it.next();
-        nodeDepths.get(currentNodes).depth = depth;
-        nextNodes.addAll(nodes.get(currentNodes).getOutlinks());
-      }
-      currentNode = nextNodes;
-      depth++;
-    }
+    int depth = calcNodeDepths(root, nodes, nodeDepths);
     ArrayList<ArrayList<Node>> res = new ArrayList<>();
     for (int i = 0; i < depth; i++) {
       res.add(new ArrayList<>());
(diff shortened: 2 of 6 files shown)


FarisElghlan:
Didn't you set it to 25 and not to 30?
`+        <module name="MethodLength">`
`+            <property name="max" value="25"/>`
`+        </module>`

I think we should wait with merging this till after the deadline of the 2nd sprint (so till tomorrow). This will probably cause more checkstyle warnings and we don't have much time left.

wouterjsmit:
Agreed. And yes, I will fix that to correctly specify `30`. Shows just how much I doubted.

wouterjsmit:
With the now passing build, this amendment reveals some new errors. Will fix these tonight.

wouterjsmit:
I suggest that the authors of the violating methods fix them themselves. As is the problem with long methods, they are incomprehensible.

There are 3 violations:
the Sniptest in PL2-Core is too long (removing blank lines fixes this, but the test should be rewritten to use the setup and teardown interface in JUnit instead, as well as actually test anything?)

Some method in DrawGraph contains commented code which pushes it over the limit.

onChange in NodeDensityHeatmap is 1 line too long. It seems to contain some obvious blocks that could be abstracted away. Removing one blank line (that separates the blocks) works, but is not desired.

Please push changes to these methods to this branch, so we can merge this, @cathmer @FarisElghlan .

**Also, please provide some input on the multiple variable declaration rule.**

FarisElghlan:
Uh it is still set to 25?

FarisElghlan:
By the way I didn't say 30 is better than 25. I actually think 30 is a bit too large, but 20 is a bit small, so I think 25 could actually work pretty well.

wouterjsmit:
Oh, you're right. I didn't push that change. Regardless of the number
though, those methods need to be split (and assuming 25, others need too)

wouterjsmit:
@cathmer @FarisElghlan @cbilstra Could you look at your violations?

FarisElghlan:
Fixed the checkstyle errors (might cause some merge conflicts with current pull requests though).

PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/FilterSnipsTest.java
```diff
@@ -1,51 +1,12 @@
...
-import nl.tudelft.pl2016gr2.model.Node;
-import nl.tudelft.pl2016gr2.model.OriginalGraph;
 
 import org.junit.Test;
 
 public class FilterSnipsTest {
 
   @Test
   public void testMakeSnip() {
-    OriginalGraph graph = new OriginalGraph();
```
wouterjsmit: We might as well remove the whole file now then

FarisElghlan: I tried to avoid merge conflicts with pull request #52. This deletion will probably cause one (or two) merge conflicts, but deleting the file might cause more merge conflicts (I am not really sure about this though).

> Pull request #52 (Test/filter-snips) adds tests to this test class

wouterjsmit: I see, Since this violating hunk of code is outdated with #52, there is no need to correct it here. We can update it with `dev` when #52 is merged.

edit: Not that I am not thankful that you did fix these errors, btw

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawGraph.java
```diff
@@ -87,36 +86,35 @@ private void drawEdges(Pane pane, HashMap<Integer, Node> nodes,
...
-        //          bendCurve(curve, fromCircle, toCircle);
-        //        });
-        //
-        //        pane.getChildren().add(curve);
-        //        curve.toBack();
       }
     });
   }
 
+  //        QuadCurve curve = new QuadCurve();
```
wouterjsmit: I tried to use `git blame` to find out who wrote these lines initially, but I can only seem to get @FarisElghlan as author.

Do we really need to save this snippet in comments?

wouterjsmit:
Overall great improvement @FarisElghlan.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawGraph.java
FarisElghlan: This was added during pull request #27. I didn't really agree with merging this code, since it is not usefull yet (as I had also stated in my last comment on that pull request), since the lines were just bend arbitrarily and didn't avoid collision, so I just commented the code (and moved it out of the method to make the method smaller, which maybe caused the git blame to be incorrect?).

wouterjsmit: Thomas quite specifically asked for curved lines, so any curvature is
better than none I would say

cathmer:
@Pathemeous Why is this branch not being merged in dev?

wouterjsmit:
@cathmer without looking at code, I think it's the modifications that were made in #52 or #61. Will resolve these myself

PR #37, opened 2016-05-05
Feat/extend parser

cbilstra:
Extended the parser, such that bases are loaded into nodes, together with genomes which 'visit' nodes. Could be improved by using genome indexes instead of Strings. Parser package has 100% test-coverage.


FarisElghlan:
The tests should be put into the test package and not in the source package (check the shared package for an example).
Also please merge dev into your branch and fix all of the checkstyle/pmd errors.

FarisElghlan:
With 'all of the checkstyle/pmd errors' I only meant the errors in the module you are currently working in by the way (so just the parser module).

FarisElghlan:
Fixed these issues myself, since Cas is unavailable at the moment. This branch is ready for merging.

FarisElghlan:
Please choose more meaningful names for private method and comments for private methods during the next sprint.
In the `FullGfaReader` class for example, the private `handleH`, `handleS` and `handleL` methods are undocumented and their names don't say much either. The only thing I know is that `handleH` for example is called when an 'H' is read and I was only able to derive that because I remember the file format. If the method `handleH` for example reads an edge from the file, call it something like `parseEdge` for example.

FarisElghlan:
Fixed all of the merge conflicts caused by other pull requests.

wouterjsmit:
Great! We now have a working build!

PR #50, opened 2016-05-07
Add console output for PMD errors during build process

wouterjsmit:
Also adds tests to Findbugs inspection, since why not.


FarisElghlan:
Does this really have to be reviewed? If you've tested that it works correctly you may merge it.

PR #63, opened 2016-05-10
Test/model

cbilstra:
Tested some classes of the model in the shared-package. Also had to do some checkstyle adjustments in the annotation utility we use.

### DIFF
--- a/PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
+++ b/PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
@@ -102,11 +102,11 @@ public class FullGfaReader {
     //words[4] = CRDCTG:Z:XXXX
     //words[5] = CTG:Z:XXXX;XXXX;...
     //words[6] = START:Z:INT
-    Node no = originalGraph.getNode(nodeId);
-    no.setGenomes(nodeGens);
-    no.setBases(bases);
-    no.setSequenceLength(bases.length());
-    no.setAlignment(orientation);
+    Node node = originalGraph.getNode(nodeId);
+    node.setGenomes(nodeGens);
+    node.setBases(bases);
+    node.setSequenceLength(bases.length());
+    node.setAlignment(orientation);
   }
 
   /**
@@ -114,13 +114,13 @@ public class FullGfaReader {
    * @param sc The scanner which is used to parse the file.
    */
   private void handleL(Scanner sc) {
-    int parent = sc.nextInt();
+    int parentId = sc.nextInt();
     sc.next();
-    int child = sc.nextInt();
-    Node par = originalGraph.getNode(parent);
-    Node ch = originalGraph.getNode(child);
-    par.addOutlink(ch.getId());
-    ch.addInlink(par.getId());
+    int childId = sc.nextInt();
+    Node parent = originalGraph.getNode(parentId);
+    Node child = originalGraph.getNode(childId);
+    parent.addOutlink(child.getId());
+    child.addInlink(parent.getId());
     sc.nextLine();
   }
   
--- a/PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
+++ b/PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
@@ -133,8 +133,7 @@ public class FullGfaReader {
     String[] words = line.split(":");
     if (words[0].endsWith("ORI")) {
       String[] gens = words[2].split(";");
-      ArrayList<String> gensAr = new ArrayList<>();
-      gensAr.addAll(Arrays.asList(gens));
+      ArrayList<String> gensAr = new ArrayList<>(Arrays.asList(gens));
       return gensAr;
     }
     return null;
--- a/PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
+++ b/PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
@@ -50,7 +50,7 @@ public class FullGfaReader {
       graph.addNode(new Node(i, 1, new ArrayList<>(), 0));
     }
   }
-
+  
   /**
    * Read the actual data.
    */
@@ -77,35 +77,42 @@ public class FullGfaReader {
     System.out.println("Number of lines read: " + count);
   }
 
+  /**
+   * Private method for handling an "H" as first character on the line of a gfa file.
+   * @param sc The scanner which is used to parse the file.
+   */
   private void handleH(Scanner sc) {
     String line = sc.nextLine();
-    this.genoms = extractGenoms(line);
+    this.genoms = extractGenomes(line);
 
   }
 
-  private ArrayList<String> extractGenoms(String line) {
-    String[] words = line.split(":");
-    if (words[0].endsWith("ORI")) {
-      String[] gens = words[2].split(";");
-      ArrayList<String> gensAr = new ArrayList<>();
-      gensAr.addAll(Arrays.asList(gens));
-      return gensAr;
-    }
-    return null;
-  }
-
+  /**
+   * Private method for handling an "S" as first character on the line of a gfa file.
+   * @param sc The scanner which is used to parse the file.
+   */
   private void handleS(Scanner sc) {
     int nodeId = sc.nextInt();
     String bases = sc.next();
     String line = sc.nextLine();
     String[] words = line.split("\\s+");
-    ArrayList<String> nodeGens = extractGenoms(words[2]);
+    ArrayList<String> nodeGens = extractGenomes(words[2]);
+    int orientation = extractStart(words[6]);
+    //words[3] = CRD:Z:XXXX
+    //words[4] = CRDCTG:Z:XXXX
+    //words[5] = CTG:Z:XXXX;XXXX;...
+    //words[6] = START:Z:INT
     Node no = originalGraph.getNode(nodeId);
     no.setGenomes(nodeGens);
     no.setBases(bases);
     no.setSequenceLength(bases.length());
+    no.setAlignment(orientation);
   }
 
+  /**
+   * Private method for handling an "L" as first character on the line of a gfa file.
+   * @param sc The scanner which is used to parse the file.
+   */
   private void handleL(Scanner sc) {
     int parent = sc.nextInt();
     sc.next();
@@ -115,6 +122,35 @@ public class FullGfaReader {
     par.addOutlink(ch.getId());
     ch.addInlink(par.getId());
     sc.nextLine();
+  }
+  
+  /**
+   * This method extracts the single genomes from the ORI tag.
+   * @param line the full ORI tag
+   * @return an arrayList of the names of the genomes.
+   */
+  private ArrayList<String> extractGenomes(String line) {
+    String[] words = line.split(":");
+    if (words[0].endsWith("ORI")) {
+      String[] gens = words[2].split(";");
+      ArrayList<String> gensAr = new ArrayList<>();
+      gensAr.addAll(Arrays.asList(gens));
+      return gensAr;
+    }
+    return null;
+  }
+  
+  /**
+   * This method extracts the alignment integer from the START tag.
+   * @param line the full START tag
+   * @return the alignment integer of the tag
+   */
+  private int extractStart(String line) {
+    String[] words = line.split(":");
+    if (words[0].endsWith("START")) {
+      return Integer.parseInt(words[2]);
+    }
+    return 0;
   }
 
   /**
--- a/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
+++ b/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
@@ -17,10 +17,14 @@ public class OriginalGraph implements GraphInterface {
   /**
    * Print a string representation of this graph.
    */
-  public void print() {
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
     for (Node node : nodes.values()) {
-      System.out.println(node);
+      sb.append(node);
+      sb.append("\n");
     }
+    return sb.toString();
   }
 
   @Override
--- a/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
+++ b/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
@@ -21,8 +21,7 @@ public class OriginalGraph implements GraphInterface {
   public String toString() {
     StringBuilder sb = new StringBuilder();
     for (Node node : nodes.values()) {
-      sb.append(node);
-      sb.append("\n");
+      sb.append(node).append('\n');
     }
     return sb.toString();
   }
--- a/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/utility/AccessPrivate.java
+++ b/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/utility/AccessPrivate.java
@@ -51,7 +51,8 @@ public final class AccessPrivate {
     } catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException ex) {
       Logger.getLogger(AccessPrivate.class.getName()).log(Level.SEVERE, null, ex);
     }
-    throw new TestAnnotationException("An error occured while accessing the" + " method with test id: " + id + ".");
+    throw new TestAnnotationException("An error occured while accessing the" 
+        + " method with test id: " + id + ".");
   }
 
   /**
@@ -121,7 +122,8 @@ public final class AccessPrivate {
     try {
       Field field = AccessPrivate.getField(cl, id);
       if (Modifier.isFinal(field.getModifiers())) {
-        throw new TestAnnotationException("Trying to set the value of " + "the final field with id: " + id + ".");
+        throw new TestAnnotationException("Trying to set the value of " 
+            + "the final field with id: " + id + ".");
       }
       field.setAccessible(true);
       field.set(obj, value);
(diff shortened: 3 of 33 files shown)


wouterjsmit:
Why exactly did you have to make adjustments to the annotation classes?
They built just fine before. Am a bit hesitant to modify 3rd party code.

Will have a look at the rest of the code later

cbilstra:
Test exceeded the max length of a line for example.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -77,34 +77,41 @@ private void read() {
...
+    ArrayList<String> nodeGens = extractGenomes(words[2]);
+    int orientation = extractStart(words[6]);
+    //words[3] = CRD:Z:XXXX
+    //words[4] = CRDCTG:Z:XXXX
+    //words[5] = CTG:Z:XXXX;XXXX;...
+    //words[6] = START:Z:INT
     Node no = originalGraph.getNode(nodeId);
     no.setGenomes(nodeGens);
     no.setBases(bases);
+    no.setAlignment(orientation);
```
FarisElghlan: Please use `node` as a variable name instead of `no`. `no` looks strange and `node` isn't much longer.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -115,6 +122,35 @@ private void handleL(Scanner sc) {
...
+   * @param line the full ORI tag
+   * @return an arrayList of the names of the genomes.
+   */
+  private ArrayList<String> extractGenomes(String line) {
+    String[] words = line.split(":");
+    if (words[0].endsWith("ORI")) {
+      String[] gens = words[2].split(";");
+      ArrayList<String> gensAr = new ArrayList<>();
+      gensAr.addAll(Arrays.asList(gens));
+      return gensAr;
```
FarisElghlan: You can instantiate an arraylist with a collection as argument:
<code>return new ArrayList<>(Arrays.asList(gens));</code>

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
```diff
@@ -17,10 +17,14 @@ public OriginalGraph() {
...
    * Print a string representation of this graph.
    */
-  public void print() {
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
     for (Node node : nodes.values()) {
-      System.out.println(node);
+      sb.append(node);
+      sb.append("\n");
```
FarisElghlan: Appending '\n' as character is more efficient (not really important for a toString, but usually StringBuilder is used for efficiency, so just some advice).
Also in my opinion chaining the calls to stringbuilder looks nicer:
<code>sb.append(node).append('\n');</code>

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/BubbledGraphTest.java
```diff
@@ -0,0 +1,21 @@
...
+   * Setup, creates a new graph.
+   */
+  @Override
+  public GraphInterface getInstance() {
+    //ArrayList<Bubble> bubbles = new ArrayList<Bubble>();
+    Bubble b2 = new Bubble(2, 4);
+    Bubble b1 = new Bubble(5, 8);
+    //bubbles.add(b1);
+    //bubbles.add(b2);    
+    GraphInterface gra = new BubbledGraph();
```
FarisElghlan: Remove the commented code?

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/GraphTest.java
```diff
@@ -0,0 +1,34 @@
+package nl.tudelft.pl2016gr2.model;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public abstract class GraphTest {
+  
+  public GraphInterface gra;
```
FarisElghlan: Don't needlessly use very short illogical variable names (i.e. use `graph` instead of `gra` in this case).

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/OriginalGraphTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+import java.util.ArrayList;
+
+public class OriginalGraphTest extends GraphTest {
+  
+  private OriginalGraph curGra;
```
FarisElghlan: `curGraph`

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/MetaDataTest.java
```diff
@@ -0,0 +1,42 @@
+package nl.tudelft.pl2016gr2.model;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class MetaDataTest {
+  
+  private MetaData md;
```
FarisElghlan, line 10: `metaData`

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/OriginalGraphTest.java
```diff
@@ -0,0 +1,49 @@
...
+    gra.addNode(n1);
+    gra.addNode(n2);
+    gra.setGenoms(ar);
+    this.curGra = gra;
+    return gra;
+  }
+  
+  @Test
+  public void testGetNodes() {
+    assertEquals(curGra.getNodes().size(), 2);
```
FarisElghlan: The first parameter of assertEquals should be the expected value and the second parameter should be the actual value:
<code>assertEquals(2, curGra.getNodes().size());</code>

This applies to pretty much all other assertEquals statements in your code too.

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/OriginalGraphTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.ArrayList;
+
+public class OriginalGraphTest extends GraphTest {
+  
+  private OriginalGraph curGra;
+
+  @Override
+  public GraphInterface getInstance() {
+    OriginalGraph gra = new OriginalGraph();
```
FarisElghlan: Local variable <code>gra</code> hides the public <code>gra</code> field of the parent class.

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/NodeTest.java
```diff
@@ -0,0 +1,66 @@
...
+  
+  @Test
+  public void basesTest() {
+    node.setBases(bases);
+    assertEquals(node.getBases(), "actcaggcagatcattcgctagacat");
+  }
+  
+  @Test
+  public void genomeTest() {
+    assertEquals(node.getGenomes().get(1).toString(), "ref2.fasta");
```
FarisElghlan: You're calling <code>toString()</code> on a <code>String</code>. This is redundant

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/BubbledGraphTest.java
```diff
@@ -0,0 +1,21 @@
+package nl.tudelft.pl2016gr2.model;
+
+public class BubbledGraphTest extends GraphTest {
+  
+  /**
+   * Setup, creates a new graph.
+   */
```
FarisElghlan: Missing <code>@return</code> tag in comments.

FarisElghlan: If you write comments in the parent class you may leave out the comments for overwritten methods in the child class (if no additional description is needed). The child class will simply inherit the comments from the parent class.

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/BubbledGraphTest.java
FarisElghlan: Local variable <code>gra</code> hides the public <code>gra</code> field of the parent class.

FarisElghlan:
Reviewed the code (see line comments in the `files changed` section). In general please try to use more descriptive variable names and don't use shorter versions of variable names if the full name isn't long (ex. don't use <code>gra</code> instead of <code>graph</code>, or <code>no</code> instead of <code>node</code>).
Since java 1.8 you may also leave out the class name during the instantiation of a generic class (i.e. use diamond operator) or completely leave out the generic class name:
You can write
`ArrayList<String> stringList = new ArrayList<String>();`
as
`ArrayList<String> stringList = new ArrayList<>();`
or
`ArrayList<String> stringList = new ArrayList();`
This is however a matter of preference, so I didn't comment on it in the code.

wouterjsmit:
@cbilstra Regarding the checkstyle in the annotations classes, I saw that you changed the package naming in 814f207, which caused the builds to fail (on those classes). In the future, if you make a commit, try building it locally. This avoids the case taht when you change something, a second commit has to fix the style warnings for the first commit (which I will ask you to squash, so in the end it saves you effort)

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/utility/AccessPrivate.java
```diff
@@ -1,4 +1,4 @@
-package nl.tudelft.pl2016gr2.test.utility;
```
wouterjsmit: Your commit where you change the package naming does not contain a description. Because of that, I do not know what your reason for this rename was.

What is wrong with it? In that layer (the pl2016gr2 folder) it is normal to specify subfolders based on what its contents are for (these utilities are for tests only)

cbilstra: I changed it because I found it very confusing that a package containing 'test' is not in src/test but in src/main. Checkstyle had difficulties with that class so I tried to find it which was confusing because I was first looking in src/test. Maybe we should move the utility package to src/test.

cbilstra: About the checkstyle in the annotations classes, they causes the build to fail because of a line length bigger than 100 characters, not because I changed the package name. Furthermore, I built it locally and the build succeeded.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
cbilstra: Agree. Will change this in another PR though because it is not really part of the tests. Needed to add the alignment for testing purpposes.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
cbilstra: I never use this syntax myself, I guess somebody changed this?

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
cbilstra: My opinion is that chaining actually makes it less readable. As \n creates a new line, I also put it on a new line.

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/BubbledGraphTest.java
cbilstra: Agree with the commented code. About gra: This is not a problem because gra in the parent class is set using getInstance(). I gave it the same name so it would be clearer that they have to do with the same variable.

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/MetaDataTest.java
cbilstra, line 10: I do not agree with this one, in general I'm not using long variable names and metaData is much longer than md.

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/OriginalGraphTest.java
cbilstra: This is not a problem because gra in the parent class is set using getInstance(). I gave it the same name so it would be clearer that they have to do with the same variable.

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/NodeTest.java
cbilstra: Wow indeed

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/BubbledGraphTest.java
cbilstra: I will add javadoc in the parent class.

wouterjsmit: > Missing @return tag in comments.

I opened an issue about this: #68

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
wouterjsmit: @cbilstra I rather it is changed now. It is a minor change and will probably be forgotten. We'll see it as 'refactoring the code somewhat and testing it' rather than only 'testing it'. A.k.a. Cleanup

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
wouterjsmit: @FarisElghlan not to be very nitpicky, but `sb.append(node + '\n')` would just be a concatenation inside the argument, wouldn't it?

@cbilstra for characters that would normally appear on a newline I would agree, but the newline character in text only appears at the end of a line (so directly after node). If you had another line of text appended, then that would as you say, fall onto the next line.

```
sb.append(node).append('\n');
sb.append("next line of text");
```

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/utility/AccessPrivate.java
wouterjsmit: @cbilstra we should definitely not move it to `src/test` because that is where the test classes reside.

We could consider a different folder name, because I agree it can be somewhat confusing.
I would like the folder structure to have the distinction of:
1) these classes are not ours, they're 3rd party
2) these classes are utilities for **testing** only.

My suggestions:
`nl.tudelft.pl2016gr2.thirdparty.testing.utility`
`nl.tudelft.pl2016gr2.thirdparty.testingutility`

`testing` is different from `test` (it is for 'testing', rather than it is a 'test')

wouterjsmit:
@cbilstra Please incorporate the extra checkstyle rules from #36 in this branch (and resolve conflicts). If ther eare conflicts from the changes in #36, you can obviously discard those and keep your own code (but do make sure it is checkstyle-compliant)

Best way:
`git fetch origin`
`git rebase origin/dev`

Other way:
merging as per usual

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
FarisElghlan: @Pathemeous I didn't mean "appending '\n' as character" as in appending it to the string directly. I meant appending it to the StringBuilder as a character instead of appending a String (i.e. calling `Stringbuilder.append(char)` instead of `StringBuildren.append(String)`)
`sb.append(node + '\n')` would indeed be less efficient, but `sb.append(node); sb.append('\n')` is more efficient
this isn't really that important though

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/BubbledGraphTest.java
wouterjsmit: @cbilstra @FarisElghlan I believe you do need to insert `{@inheritDoc}`, as found here http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/MetaDataTest.java
FarisElghlan, line 10: It is indeed longer, but not that long. `md` doesn't describe the variable in any meaningful way at all and with the easy of use of code completion in IDEs it doesn't even take longer to "type" a 30 character variable name than a 5 character variable name. Using very short variable names in general is in my opinion a bad practise and not a good argument to use short variable names.

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/BubbledGraphTest.java
FarisElghlan: `{@inheritDoc}` is only needed for explicit comment inheritance, i.e. if you want to inherit the documentation and want to add extra details to it. For comment inheritance without any changes (exact copy of the parent class' comments) you can just leave out the comment section completely.

wouterjsmit: Can't test this on our project since apparently all inheritances (that I could find) lack javadoc, but disregard my comment then

cbilstra:
Processed all feedback, ready for merge with dev

cbilstra:
Also squashed the "fix ..." commits. Wanted the final commit message to be different but that did not go as intended.

wouterjsmit:
Thanks, merging.

PR #64, opened 2016-05-10
Add the 'test' group to the contributing document.

cbilstra:
Could not directly push to dev, so have to do it with a PR. Only 2 lines have changed so this can be merged right away.


wouterjsmit:
I was just thinking about adding this. Merging

PR #69, opened 2016-05-11
Filter bubbles

cathmer:
In this PR a class is created which can be used to filter bubbles based on the phylogenetic tree. What it does right now, is it takes the graph and the phylogenetic tree, and creates a new graph which contains only nodes through which all genomes of the phylo tree run, and it adds mutation bubbles in between these shared nodes.

The filter bubbles have also been tested.

There are two major issues which remain towards semantic zooming:
- The customer really wants a top level view.Right now, the filtered graph (of only 10 genomes) still contains 420 nodes, which is too much to display on a single screen. These nodes are in a straight line though, so they can be easily merged together into bigger nodes, but this should be done in such a way that the bigger nodes still have biological relevance. 
- It should be possible to zoom in the graph. This is the first problem I will be addressing. Because from the filtered graph, it should be possible to zoom back in.


wouterjsmit:
@cathmer I just merged #36. this poses some more checkstyle rules, which you should test your branch against.

Also, apparently the checkstyle fixes there do conflict with your branch.

Assuming your version is checkstyle-correct, you should be easily able to discard those changes and keep your own.

PL2/PL2-core/pom.xml
```diff
@@ -26,6 +26,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-all</artifactId>
+			<version>1.9.5</version>
+		</dependency>
+        <dependency>
```
FarisElghlan, line 33: In the GUI module I am using this dependensy for mockito:

``` xml
<dependency>
            <groupId>org.mockito</groupId>
            <artifactId>mockito-core</artifactId>
            <version>1.10.19</version>
        </dependency>
</dependency>
```

@Pathemeous @justinvdk which mockito version is preferred? It would be nicer to use the same version (then travis/anyone building the project for the first time won't have to download both either).

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
```diff
@@ -0,0 +1,149 @@
...
+   * tree based on which this graph will be filtered.
+   * @param originalGraph : the graph to be filtered.
+   * @param root : the node of the phylogenetic tree to use for filtering. This will be the root of 
+   *     the tree when a full overview is wanted.
+   */
+  public FilterBubbles(OriginalGraph originalGraph, IPhylogeneticTreeNode root) {
+    //this.graph = graph;
+    this.originalGraph = originalGraph;
+    this.treeRoot = root;
+    mutationId = 10000;
```
FarisElghlan, line 42: Is this a hardcoded value for a unique ID, which will break when we load a graph with more than 10 000 nodes? If this is the case you should store a "highestId" value in the OriginalGraph class and create a new unique ID based on that, or maybe use negative IDs for bubbles and positive IDs for nodes (not sure if this will work, as I didn't write this part of the code, but this could be a nice workaround).

Also remove the commented graph variable if it isn't needed.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
```diff
@@ -6,17 +6,23 @@
 public class OriginalGraph implements GraphInterface {
 
   private HashMap<Integer, Node> nodes;
+  private HashMap<Integer, AbstractNode> abstractNodes;
```
FarisElghlan, line 9: Why did you add this variable? It seems like an exact copy of variable `nodes`, but with different generics (since you call `put()` on both of them each time you add anything to either of them).
If this variable is really only needed for `getAbstractNodes()` and contains the same objects as `nodes`, then I suggest you delete `abstractNodes` and change the method `getAbstractNodes()` to:

``` java
  @Override
  public HashMap<Integer, ? extends AbstractNode> getAbstractNodes() {
    return nodes;
  }
```

And don't forget to change the method in the interface to:

``` java
HashMap<Integer, ? extends AbstractNode> getAbstractNodes();
```

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/GraphInterface.java
```diff
@@ -10,5 +14,5 @@
 
   AbstractNode getRoot();
 
-  // public HashMap<Integer, AbstractNode> getNodes();
+  HashMap<Integer, AbstractNode> getAbstractNodes();
```
FarisElghlan, line 17: Add comments to the methods in this interface? (then the comments will be automatically inherited by classes which implement this interface, so you will only have to write them for the parent class, unless if you want to add extra details for a specific implementation of the interface of course)

FarisElghlan:
Reviewed the code. Looks good overall, added a few line comments on minor issues.

PL2/PL2-core/pom.xml
wouterjsmit, line 33: We will use the current highest release, which we will keep for the entire project unless some future release is required for certain functionality.

currently, the highest release version is `1.10.19`, so we will use that (see [Maven Repos](http://mvnrepository.com/artifact/org.mockito/mockito-all)

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
wouterjsmit, line 42: The `mutationId` is also only used in `FilterBubbles()`. Not sure what the idea is behind it but you could also consider making it a local variable instead.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.LinkedList;
+import java.util.PriorityQueue;
+import java.util.Queue;
+
+/**
+ * Class used to filter bubbles based on the phylogenetic tree. 
+ * @author Casper
+ *
+ */
+public class FilterBubbles {
```
wouterjsmit, line 23: You might want to consider to more elaborately explain the functionality of this class (at the class JavaDoc or the constructor's) Reading either of these right now still doesn't give me a clear idea of what exactly it does or what the bubbles are based on exactly, even though I'm supposed to know it because we discussed it.

Edit: It is also hard to review the structure of this class since I don't exactly know what it is it does

```
 * Creates a FilterBubbles object, with the graph to be filtered and the node of the phylogenetic
 * tree based on which this graph will be filtered.
```

I don't really get how the filter is based on a certain node in the phylo tree, for instance. What semantic goal is this linked to? Showing differences between different evolution branches more clearly?

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
```diff
@@ -0,0 +1,149 @@
...
+      bubble.addInlink(node.getId());
+      bubble.addOutlink(end.getId());
+      
+      filteredGraph.addNode(node);
+      filteredGraph.addNode(bubble);
+      filteredGraph.addNode(end);
+      
+      previous = next;
+      mutationId++;
+    }
```
wouterjsmit, line 95: FilterBubbles.java contains many long methods. This one will trigger the methodLength checkstyle rule.
Consider creating private methods that deal with a single block, or increasing functionality of the model classes (`Bubble` / `Node`) if there is some modifications that should be handled by the class itself.

`end.getInlinks().clear();` (line `80`) is an example of using a getter to modify the internal data. This should be abstracted in such a way that a `Node` offers a public method `clearInLinks()`, `clearOutLinks()` or even `clearAllLinks()` (even beter would be to specify this method in an interface, such as `AbstractNode`).

PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubblesTest.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.PriorityQueue;
+
+/**
+ * Tests the FilterBubbles class. The private method buildGraph is not tested,
+ * because by testing the filter method and the getSharedNodes method this 
+ * method is basically already tested. 
+ * @author Casper
+ *
+ */
+public class FilterBubblesTest {
```
wouterjsmit, line 29: You can use a JavaDoc link:
`Tests the {@link FilterBubbles} class. (...)`

Exact syntax I always have to look up, but quite certain this one is correct

PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubblesTest.java
```diff
@@ -0,0 +1,102 @@
...
+   */
+  @Before
+  public void setup() {
+    leaves = new ArrayList<>();
+    for (int i = 1; i <= 5; i++) {
+      leaves.add("leaf" + i);
+    }
+    
+    when(treeRoot.getLeaves()).thenReturn(leaves);
+    originalGraph = new FullGfaReader("FilterBubblesTest.gfa", 11).getGraph();
```
wouterjsmit, line 48: Maybe consider mocking an `OriginalGraph`, since it is a dependency of the class.

PL2/PL2-launcher/src/main/java/nl/tudelft/pl2016gr2/launcher/Dnav.java
```diff
@@ -57,11 +57,12 @@ private void insertData(RootLayoutController controller) {
...
     // abusing NWKReader class as this class' classloader can access the correct resource
     Reader reader = new InputStreamReader(
-        FullGfaReader.class.getClassLoader().getResourceAsStream("340tree.rooted.TKK.nwk"));
+        FullGfaReader.class.getClassLoader().getResourceAsStream("10tree_custom.rooted.TKK.nwk"));
     BufferedReader br = new BufferedReader(reader);
     TreeParser tp = new TreeParser(br);
 
-    tree = tp.tokenize("340tree.rooted.TKK");
+    tree = tp.tokenize("10tree_custom.rooted.TKK");
+
```
wouterjsmit, line 65: I'm not sure we should keep committing when we switch between data sets. these lines should be kept as is. Could you unstage them from the commit that added them?

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
wouterjsmit, line 9: I would like to add that if the `OriginalGraph` uses `Node`, it should override, using covariance, the return types of the `GraphInterface`. The interface should not specify two types of nodes, and its implementations should not offer two types of the same inheritance.

If OriginalGraph doesn't necessarily use `Node`, but you want an OriginalGraph that does, simply extend it.

cathmer:
I branched off this branch and edited the filter algorithm, so this pull request is outdated. Therefore I will close it.
However, I will keep in mind the review of this PR.

PR #73, opened 2016-05-12
Create Split Graphs Algorithm

### LINKED ISSUE #24: Split graph in subgraphs to show in visualizer
- [ ] Subgraph nodes should be a new object instantiation rather than a reference to the same object (objects should be copied)
- [ ] Nodes in the subgraph, even though they are copies, should use the exact same ID as in the main graph
- [ ] More specifically, the only thing that may differ for these nodes is the `genome list`, the `inLinks` and the `outLinks`, because some elements of these lists may not be present in the subgraph.
- [ ] The algorithm will require an `OriginalGraph` and a list of genomes that should be present in the subgraph. It returns a new instantiation of `OriginalGraph` that only contains nodes for the specified genomes, with only links between these nodes.

wouterjsmit:
Closes #24.

Currently still have to add tests for both the algorithm itself and the extension in the model (genome iterator). However, I am waiting with the model test for #63 to be merged.

The main code is done however and is ready for review.

### DIFF
--- a/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
+++ b/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
@@ -29,7 +29,7 @@ import java.util.NoSuchElementException;
 public class SplitGraphs {
 
   @TestId(id = "graph_field")
-  private OriginalGraph mainGraph;
+  private final OriginalGraph mainGraph;
 
   /**
    * Instantiates an algorithmic class on a <code>OriginalGraph</code>.
@@ -57,7 +57,7 @@ public class SplitGraphs {
    */
   public OriginalGraph getSubgraph(Collection<String> genomes) {
     if (!mainGraph.getGenoms().containsAll(genomes)) {
-      throw new NoSuchElementException("All genomes must present in the main graph.");
+      throw new NoSuchElementException("All genomes must be present in the main graph.");
     }
     OriginalGraph subGraph = buildSubgraph(genomes);
     return prune(subGraph, genomes);
--- a/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
+++ b/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
@@ -2,6 +2,7 @@ package nl.tudelft.pl2016gr2.core.algorithms;
 
 import nl.tudelft.pl2016gr2.model.Node;
 import nl.tudelft.pl2016gr2.model.OriginalGraph;
+import nl.tudelft.pl2016gr2.test.utility.TestId;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -27,6 +28,7 @@ import java.util.NoSuchElementException;
  */
 public class SplitGraphs {
 
+  @TestId(id = "graph_field")
   private OriginalGraph mainGraph;
 
   /**
--- a/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
+++ b/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
@@ -3,6 +3,7 @@ package nl.tudelft.pl2016gr2.core.algorithms;
 import nl.tudelft.pl2016gr2.model.Node;
 import nl.tudelft.pl2016gr2.model.OriginalGraph;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -57,7 +58,7 @@ public class SplitGraphs {
       throw new NoSuchElementException("All genomes must present in the main graph.");
     }
     OriginalGraph subGraph = buildSubgraph(genomes);
-    return prune(subGraph);
+    return prune(subGraph, genomes);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/collectioninterfaces/GenomeIterator.java
```diff
@@ -0,0 +1,105 @@
...
+    Queue<Node> queue = new LinkedList<>();
+    queue.add(graph.getRoot());
+
+    while (!queue.isEmpty()) {
+      Node current = queue.remove();
+      if (current.getGenomes().contains(genome)) {
+        return current;
+      }
+
+      queue.addAll(graph.getTargets(current.getId()));
```
cathmer, line 101: I would keep track of nodes that are already visited and of nodes that are already in the queue. Otherwise you will visit the same node multiple times, which is inefficient. So I would make a HashSet (or something similar) to keep track of the visited nodes, and also check if a node is not already in the queue beforea adding it.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/collectioninterfaces/GenomeIterator.java
```diff
@@ -0,0 +1,105 @@
...
+   * NoSuchElementException} will be thrown.
+   * </p>
+   *
+   * @param graph  The graph to traverse through
+   * @param genome The genome to find and iterate over
+   */
+  public GenomeIterator(OriginalGraph graph, String genome) {
+    this.genome = genome;
+    this.graph = graph;
+    this.current = findFirstGenomeOccurrence();
```
cathmer, line 44: It's a little confusing to the user of this class to get the root in the constructor. You can, for example, instead ad an if statement to hasNext() to check whether current == null. If this is the case, you find the root, otherwise you just run the usual code.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
```diff
@@ -0,0 +1,151 @@
...
+   * Instantiates an algorithmic class on a <code>OriginalGraph</code>.
+   * <p>
+   * When {@link #getSubgraph(Collection)} is called, the class will construct a new instance of
+   * the
+   * graph containing only the specified genomes.
+   * </p>
+   *
+   * @param mainGraph The graph to split
+   */
+  public SplitGraphs(OriginalGraph mainGraph) {
```
cathmer, line 44: Eventually we'll want to split any graph, not just the original graph. So it might be better to insert the GraphInterface in here.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
```diff
@@ -0,0 +1,151 @@
...
+  private OriginalGraph buildSubgraph(Collection<String> genomes) {
+    OriginalGraph subGraph = new OriginalGraph();
+    // Add all nodes for every genome
+    for (String genome : genomes) {
+      Iterator<Node> iterator = mainGraph.iterator(genome);
+      // add genome root
+      subGraph.addNode(iterator.next());
+
+      while (iterator.hasNext()) {
+        subGraph.addNode(iterator.next());
```
cathmer, line 84: You could create a new node here and add it to subGraph. You can then immediately add the current genome to this node, this is more efficient then doing it afterwards. 
You can then add a check to see if the subgraph already contains a node: if this is the case, get the existing node and add the genome. Else, create a new node and add it to the subgraph.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
```diff
@@ -0,0 +1,151 @@
...
+
+  /**
+   * Builds a subset of gGenomes that is the intersection of the genomes in the graph and the node.
+   * The intersection is formally described as <code>node.getGenomes &#8745 graphGenomes</code>
+   *
+   * @param node         The node to build the subset for
+   * @param graphGenomes The set of genomes in the graph
+   * @return The intersection of the node and graph genomes
+   */
+  private ArrayList<String> pruneGenomes(Node node, Collection<String> graphGenomes) {
```
cathmer, line 142: It's probably more efficient if you create a new node already in the buildgraph method, and add the genome there (see my comment there). This method then becomes redundant.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
```diff
@@ -0,0 +1,151 @@
...
+   * </p>
+   * <p>
+   * While this method is intended solely for subgraphs, it will provide the same behaviour when
+   * called with any graph.
+   * </p>
+   *
+   * @param graph The graph to prune.
+   * @return A new <code>OriginalGraph</code> with only the relevant links.
+   */
+  private OriginalGraph prune(OriginalGraph graph, Collection<String> genomes) {
```
cathmer, line 108: If you change the class in such a way that you create new nodes in the buildsubgraph, you should edit this method in such a way that you still get the inlinks and outlinks of the nodes of the original graph.

cathmer:
The code looks good overall. It looks very neat and thought through. The code is also well-commented and well tested.

I added some comments in the code about things that could be done more efficiently and things that might be confusing to users of the class.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
wouterjsmit, line 44: I did not do so because the `GraphInterface` does not specify genome information (`AbstractNode`s do not contain genome information). OriginalGraph is the first object in the graph inheritance that provides this functionality.

cathmer:
@FarisElghlan took this branch and adapted the splitting algorithm. This PR can therefore be closed, although some of the proposed changes (mainly about the construction of the genome iterator) should still be made.

Issue #72, opened 2016-05-12
Testing document should be updated, as well as Architecture Design


wouterjsmit:
The testing document should be updated to include newly added special cases, or remove untested code that now has been tested.

Also, the testing document contains some information regarding the Architecture Design (module definition). We should update the architecture design to (also) contain this information and in general elaborate more using the current architecture.

FarisElghlan:
Will do this tomorrow, so I can include todays changes.

wouterjsmit:
@FarisElghlan Are you still able to do this?

FarisElghlan:
Was actually only referring to the testing document (didn't notice the architecture was also in the title)...
See: #79

PR #79, opened 2016-05-13
Updated the testing document

FarisElghlan:
Updated the testing document according to the tests which were added during sprint 3. The added files are the images of the testing reports.


FarisElghlan:
Fixes #72, except for the architecture design, but this didn't change much and I don't think anyone will do it.

wouterjsmit:
Looks great, screencaps are nice too!

PR #81, opened 2016-05-14
Huge performance improvement of the parser

FarisElghlan:
Improved the performance of the parser by around 2000% (20x faster).
The parser doesn't need the size of the graph anymore when it is being initialized.


FarisElghlan:
Performance:
First parsing the dataset of 10 genomes took around 2.5 seconds, now it takes around 100-120 milliseconds (20-25 times faster)
First parsing the dataset of 328 genomes took around 3 minutes, now it takes around 1.5-1.7 seconds (105 times faster)

The algorithm has a time complexity of O(n) (lineair), so I expect the dataset of 6000 genomes to be loadable within 30 seconds (6000 / 328 \* 1.7 = 31 seconds).

Simply reading all of the lines of the file into java strings takes 0.5 seconds for the dataset of 328 genomes, so there is a possible performance gain of 1.7 / 0.5 = 3.4 times the current speed. This would however only be possible if reading the data from disk and processing it happens concurrently and the data processing is faster than the data reading. If we rewrite the parser in native C code and use multithreading to process the data while it is being read I guess we could possibly make it 2-3 times faster than it currently is. I don't think this is worth the effort, unless if we get much larger datasets which take a lot longer to load (for example more than 5 minutes).
I assume we will only have to improve the performance of the parser if we get datasets of more than 60,000 genomes.

wouterjsmit:
Great work @FarisElghlan! Before I review your code, I think your comment is very valuable. I made a 'Grand Vision' issue: #82. I think it would be useful to collect all our findings and esimations in a central location such as that issue.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -3,153 +3,217 @@
...
-    System.out.println("Number of lines read: " + count);
+    index += 3;
+    int to = 0;
+    while (chars[index] != '\t') {
+      to *= 10;
+      to += chars[index++] - 48;
+    }
+    getNode(to).addInlink(from);
+    getNode(from).addOutlink(to);
   }
```
wouterjsmit:
``` java
int from = 0;
int index = 2;
(...)
```

There is a lot of arithmetic here. It might be useful to more elaborately describe this. Consider using line comments or if possible creating a constant and doing

``` java
int from = CONST_WITH_DESCRIPTIVE_NAME;
```

This way we (the rest) know what the `0` and `2` (and all other terms/factors) are based on by looking at the code.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -3,153 +3,217 @@
...
-  private void handleH(Scanner sc) {
-    String line = sc.nextLine();
-    this.genoms = extractGenomes(line);
+  private void parseNode(char[] chars) {
+    int nodeId = 0;
+    int index = 2;
+    while (chars[index] != '\t') {
+      nodeId *= 10;
+      nodeId += chars[index++] - 48;
+    }
```
wouterjsmit: Same thing about magic numbers here. Also noticing that the while loop seems identical. If the values in the loop (`10` and `48`) are independent of the method where they are used (`parseNode` or `parseEdge`), but based on some external factor, such as data structure (the char array) or the file format, this should definitely be separated from the methods. If these values have to change we will now need to do so in both methods, which is very error prone. (especially because it is such a small block).

A class-wide constant is sufficient, but a separate method is could be even better, because also keeps file length down.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -3,153 +3,217 @@
...
+    int index = curIndex;
+    index = skipTillCharacter(chars, index, '\t', 1);
+    node.setBases(new String(chars, curIndex, index - curIndex));
+    return index;
   }
 
   /**
-   * Private method for handling an "S" as first character on the line of a gfa file.
-   * @param sc The scanner which is used to parse the file.
+   * Parse the genoms of a node.
```
wouterjsmit: Please Ctrl+R on
`genoms -> genomes`

Let's stop using 'genoms' everywhere in the code. I cry everytime I have to use `getGenoms()`.
Singular: genome
Plural: genomes

https://en.wiktionary.org/wiki/genome#English

FarisElghlan: Ah.. first I was using genomes, but saw something else using genoms and assumed my version  was incorrect.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
FarisElghlan: This loop reads a number from an array of characters. 48 is the ascii code of number 0 (I could change this to '0'). *10 because when you read one more number, the previously read number must be shifted by one in base 10.
It is annoying to create a separate method for this because the index has to be updated and the nodeId has to be remembered too. This can be solved by returning a value pair from a method which contains both the ID and the new index, but this is bad for performance (an extra method call, creating variables in the method, returning values in an object which has to be created and updating the variables in the method which called it). Usually this is not an issue, but this has to be high performance code and is performed many times (once for every node which is read and the other lines which you commented on is perfomed twice for every edges).

The magic numbers +2 / +3 / etc are dependent on the format of the file:
an edge line for example looks like this:
`L  2030    +   2031    +   0M`
to skip the 2 tabs and '+' in between the two numbers I for example use `index += 3;`. I don't think there is any way to describe the meaning of 3 in this case with a variable name, as it depends on the file format. This might seem like a hacky way, it is, but this is mainly caused by the file format which we get (binary would be much better) and we can't change that.
As long as the file format doesn't change we won't have to change this code, so imo screw coding standards and lets go for performace.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
FarisElghlan: see comment below.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -151,15 +152,15 @@ private int parseNodeGenoms(Node node, char[] chars, int curIndex) {
...
   private void parseNodeOrientation(Node node, char[] chars, int curIndex) {
     int index = skipTillCharacter(chars, curIndex, '\t', 3);
     index = skipTillCharacter(chars, index, ':', 2) + 1;
     int orientation = 0;
     while (index < chars.length) {
-      orientation *= 10;
-      orientation += chars[index++] - 48;
+      orientation *= SHIFT_BY_BASE_10;
+      orientation += chars[index++] - '0';
     }
```
wouterjsmit: Ah, didn't make the connection with the ASCII value of 0. This is a lot clearer. How about putting this in a separate method? it appears 4 times in the code and it is clearly a single functionality.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -3,153 +3,218 @@
...
+          case 'L':
+            parseEdge(chars);
+            break;
+          case 'S':
+            parseNode(chars);
+            break;
+          case 'H':
+            parseHeader(chars);
+            break;
+          default:
```
wouterjsmit: Did you put in an empty default case to silence static tools?
In the default case you could log it, (maybe even throw an exception for it) because it is pretty bad if this happens.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -3,153 +3,218 @@
...
+      to += chars[index++] - '0';
+    }
+    getNode(to).addInlink(from);
+    getNode(from).addOutlink(to);
   }
 
   /**
-   * Private method for handling an "H" as first character on the line of a gfa file.
-   * @param sc The scanner which is used to parse the file.
+   * Parse a line of the GFA file which contains a node.
```
wouterjsmit: `which -> that`

' (which/that) contains a node.' is a restrictive clause. There are many lines in the GFA file, but this method only parses the ones that contain a node. It therefore restricts the definition of 'a line' in the main sentence.
Another way of looking at it: the clause is essential to the meaning of the sentence, leaving it out changes the definition (suddenly ALL, even non-node, lines will be parsed). If this is the case, you should use 'that'.

[which vs. that](http://www.getitwriteonline.com/archive/103103whichthat.htm)

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -3,153 +3,218 @@
...
+      }
+    } catch (IOException ex) {
+      Logger.getLogger(FullGfaReader.class.getName()).log(Level.SEVERE, null, ex);
     }
   }
-  
+
   /**
-   * Read the actual data.
+   * Parse a line of the GFA file which contains an edge.
```
wouterjsmit: `which -> that`
See my other comment about which vs. that.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -3,153 +3,218 @@
...
+      orientation *= SHIFT_BY_BASE_10;
+      orientation += chars[index++] - '0';
+    }
     node.setAlignment(orientation);
   }
 
   /**
-   * Private method for handling an "L" as first character on the line of a gfa file.
-   * @param sc The scanner which is used to parse the file.
+   * Parse the header part of the GFA file, which contains all of the genomes.
```
Pathemeous: This is correct usage of 'which'. the adjective clause 'which contains all of the genomes' is only informative: leaving it out retains the correct meaning of the sentence.
:)

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/GfaReader.java
```diff
@@ -28,14 +29,26 @@ public GfaReader(String filename, int graphsize) {
...
     read();
 
     sc.close();
     sc = null;
   }
 
   /**
+   * Initialize the nodes of a graph.
+   *
+   * @param graph the graph of which to initialize a set amount of node.
```
wouterjsmit: `node -> nodes`
`of -> for` (`of` implies that the nodes are already part of the graph, but must be initialized. However, you are adding new nodes to the graph, so you initialize a node `for` the graph, like a present :) )

Consider rephrasing:
`The graph to initialize a set amount of nodes for`

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
```diff
@@ -15,8 +15,17 @@ public OriginalGraph() {
   }
 
   /**
-   * Print a string representation of this graph.
+   * Construct an OriginalGraph.
```
Pathemeous: There should be a blank line in between the description and the tags. Not sure why checkstyle doesn't pick this up, but there are only limited checkstyle rules for JavaDoc

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
FarisElghlan: See the comment 2 above this one. You might have to click "Show 2 comments", because github seems to hide comments which apply to lines which have been changed after the comment was made.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
```diff
@@ -15,8 +15,17 @@ public OriginalGraph() {
...
+   * @param nodes the nodes.
+   * @param lowestId the lowest node id.
+   * @param genoms the genoms.
    */
```
wouterjsmit, line 23: There should be a blank line in between the JavaDoc description and the at-params tags. Not sure why checkstyle doesn't pick this up, but there are only limited checkstyle rules for JavaDoc

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
FarisElghlan: Mainly because there was a test case which tested that the code doesn't crash if an unknown line is read.

wouterjsmit: (imo) This really seems like something that at least should be logged and communicated to the user. If the file format is wrong the user must know that what he is looking at could be completely wrong data.
We should consider adapting the test case.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -3,163 +3,229 @@
...
-      return Integer.parseInt(words[2]);
+  private Node getNode(int id) {
+    Node node = nodes.get(id);
+    if (node == null) {
+      node = new Node(id, 1, new ArrayList<>(), 0);
+      nodes.put(id, node);
     }
-    return 0;
+    return nodes.get(id);
   }
```
wouterjsmit: You should return `node`. it is either not-null, meaning that it is the thing that you want to return, or it is null, and will be initialized in the if statement.

My original comment was to remove line `212` and `214` and condense the lines more (no need to explicitly save things you don't use more than once), but this leaves you doing 2 `get()` operations. Consider that this hashmap will be **huge** (millions of elements). The gets will most likely have some (many) collisions and will not be strictly `O(1)`.

wouterjsmit: Note: I edited my comment, as I noticed a clean performance improvement rather after I posted the original.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
```diff
@@ -3,163 +3,229 @@
...
-   * @return The graph.
+   * @return the graph.
    */
   public OriginalGraph getGraph() {
-    originalGraph.setGenoms(genoms);
+    if (originalGraph == null) {
+      originalGraph = new OriginalGraph(nodes, 1, genomes);
+    }
     return originalGraph;
   }
```
wouterjsmit: You might want to consider changing the constructor and this method. When you call the constructor now, the file will immediately be read, but not returned and you have to seperately get the result with `getGraph()`. It is more common that you initialize some reader, then call `read(...)` on it, which returns the result of reading the file. Conforming to this convention makes the code where the reader is used more obvious to others.

I see that you use a Singleton pattern here, which I don't think should be how the reader should behave (it is not in charge of storing the graph, only in reading it).

 If you really consider yours a better approach than what I suggest, please add more elaborate JavaDoc to explain the usage of the class and these key methods. For instance, tell that this method provides a Singleton object, and explain how the user should call this class in the Constructor/ Class javaDoc.

wouterjsmit:
What does the `GfaReader.java` do? I see that you changed the file, but it is not used and @cbilstra mentioned that this class could be removed.

If we remove this file, we should rename the `FullGfaReader` to simply `GfaReader`.

wouterjsmit:
`FullGfaReaderTest.java` could do with some unit tests if possible. The only test now is an end-to-end test.

wouterjsmit:
@FarisElghlan I reviewed your code.
The performance increase is great and this PR also addresses some chaos in the structure of the model package, so very much great job on that.
Most comments are style tips and grammar in JavaDoc. However, there were a few points that could be optimized further and cleaned up. Please take a moment to look at them.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/FullGfaReader.java
FarisElghlan: Yes that could be better.
This is not a singleton pattern, it has a public constructor and doesn't store a static reference to itself.
This class has to store the hasmap of nodes because it must be appended to when I try to get an ID which doesn't exist in the hasmap yet. It is not desirable for the OriginalGraph to have this behavior.
I only store the graph in a variable because like this it will return the same instance of the graph if you call getGraph multiple times (otherwise there could potentially be multiple original graphs with the same hashmap of nodes and list of genomes). This shouldn't really happen, but it is a bit more safe and doesn't cost anything.

wouterjsmit: I see the safety benefit of storing the graph rather than reading it multiple times (This is what I meant with a Singleton pattern: the graph is instantiated only once and subsequent calls return the same instance). However, maybe it would be possible to incorporate both sides the safety and usability, by letting `read()` return the `OriginalGraph` instance if it already exists, or actually 'read' it when it doesn't. This removes the need for `getGraph()` in the API, gives you the safety benefit, and makes the class behave more like a Java reader.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/GfaReader.java
FarisElghlan: Ok I will delete the file.
I actually only moved this code. It was first in the fullgfareader, but it isn't needed there anymore (it was not present in the GfaReader class because of code duplication issues) . The GfaReader class isn't used anymore, so I will delete it and rename the FullGfaReader class.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/GfaReader.java
```diff
@@ -2,98 +2,237 @@
...
-          break;
-        case "S":
-          sc.nextLine();
-          break;
-        case "L":
-          parseL();
-          break;
-        default:
-          System.out.println("Not S, H or L at: " + count);
+  private void read(String fileName) throws IOException {
```
wouterjsmit: The only thing here is that the parameter is useless. If you change the name of this method to `parse()`, it is both different from `read()` and sticks with the naming of the other methods (`parseEdge()` etc).

FarisElghlan:
Updated the code according to the feedback.

wouterjsmit:
Great. I will squash the commits into one and merge the branch

PR #84, opened 2016-05-16
Improved the use of generics in the AccessPrivate class

FarisElghlan:
It is now no longer necessary to cast the return values, as this is done with generics.
Example:
instead of :
`Color c5 = (Color)  AccessPrivate.callMethod("mapColor", NodeDensityHeatmap.class, null, 5, 10);`
you can now do:
`Color c5 =  AccessPrivate.callMethod("mapColor", NodeDensityHeatmap.class, null, 5, 10);`
or:
`Color c5 = AccessPrivate.<Color, NodeDensityHeatmap>callMethod("mapColor", NodeDensityHeatmap.class, null, 5, 10);`
This also applies to the getFieldValue method.

The types of the class and the passed object are now checked at compile time. If the given class does not correspond with the class of the given object, your code won't compile anymore.


wouterjsmit:
Great addition, thanks for the update. Merging

Issue #93, opened 2016-05-16
Test subgraph comparison


FarisElghlan:
Unit test the CompareSubgraphs and GraphOrdererThread class, which compare and align subgraphs.

FarisElghlan:
Fixed by PR #105

Issue #86, opened 2016-05-16
GUI integration tests


FarisElghlan:
The GUI has to be integration tested, so parts which can't be easily unit tested are also tested.

FarisElghlan:
Integration tests will have to be turned off on travis, because the JavaFX thread has to be initialized. This requires an "open window" command (even if no actual window is shown) and this is not possible on travis. The tests will still be performed when building the project locally.

FarisElghlan:
Fixed by PR #104

PR #94, opened 2016-05-16
Feat/gui compare two graphs

FarisElghlan:
Everything except for testing is done. Graphs can now be compared and shown to the user very quickly.

Because of the large amount of changes made in this branch I think it is a good idea to review and merge the code now and do the testing later (but still during this sprint of course). Testing will take a long time and if we wait that long there will most likely be a lot more merging issues, as some of the model classes have been modified. I have created two testing issues for the changes which were made in this branch: #86, #93


wouterjsmit:
I am nearly done working on the split graphs refactoring, which is your code but cleaned up and elaborately unit-tested.
It would be useful to review that part in its own, and then rebase this PR on dev so that at least a large chunk of these changes can be discarded in the review. 1.5K insertions is really too much to do a review on easily.

wouterjsmit:
See #95 in light of my previous comment.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/GraphNodeOrder.java
```diff
@@ -0,0 +1,65 @@
+package nl.tudelft.pl2016gr2.model;
+
+/**
+ * This class is used to decide the order of the nodes in the graph.
+ *
+ * @author Faris
+ */
+public class GraphNodeOrder implements Comparable<GraphNodeOrder> {
```
cathmer: The name of this class is not so clear. It now implies that the class contains a graph (or at least has to do with a graph) while it only applies to a node. Could you rename it to something like NodePosition.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/collectioninterfaces/IGenomeIterable.java
```diff
@@ -0,0 +1,23 @@
+//package nl.tudelft.pl2016gr2.collectioninterfaces;
```
cathmer: Why is the code in this class commented? Is it still going to be used? If not, it should be deleted.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/collectioninterfaces/GenomeIterator.java
```diff
@@ -0,0 +1,107 @@
+//package nl.tudelft.pl2016gr2.collectioninterfaces;
```
cathmer: Why is the code in this class commented? Is it still going to be used? If not, it should be deleted.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/CompareSubgraphs.java
```diff
@@ -0,0 +1,257 @@
...
+        curLevel = node.getLevel();
+      }
+      node.addPositionOffset(-emptyLevels);
+    }
+  }
+
+  /**
+   * Thread which can be used to order a subgraph and mark the overlapping nodes.
+   */
+  private static class SubGraphOrderer extends Thread {
```
cathmer, line 114: I think it's cleaner to put this class in a separate file.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/CompareSubgraphs.java
```diff
@@ -0,0 +1,257 @@
...
+      } catch (InterruptedException ex) {
+        Logger.getLogger(CompareSubgraphs.class.getName()).log(Level.SEVERE, null, ex);
+      }
+    }
+  }
+
+  /**
+   * Thread which can be used to calculate the overlapping nodes of two graph.
+   */
+  private static class OverlapThread extends Thread {
```
cathmer, line 199: also put this class in a separate file.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/GraphOrdererThread.java
```diff
@@ -0,0 +1,102 @@
...
+   * list contains the nodes in order from left to right. Nodes which are at the same horizontal
+   * position have sequential positions in the array list and have the same value for their level
+   * field.
+   *
+   * @param graph the graph.
+   * @return the node order.
+   */
+  private static HashMap<Integer, GraphNodeOrder> calculateGraphOrder(OriginalGraph graph) {
+    HashMap<Integer, GraphNodeOrder> nodeOrder = new HashMap<>();
+    HashMap<Integer, Integer> reachedCount = new HashMap<>();
```
cathmer, line 44: This method is not so easy to understand. Maybe explain a little more in detail what you're doing. Especially the use of the reachedCount variable is not very clear at first sight.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
```diff
@@ -0,0 +1,154 @@
...
+ * <li>The nodes in the subgraph will have an identical ID as their counterparts in the original
+ * graph.</li>
+ * <li>The nodes will be newly instantiated, meaning that the subgraph datastructure is not backed
+ * by the original graph.</li>
+ * </ul>
+ * </p>
+ *
+ * @author Wouter Smit
+ */
+public class SplitGraphs {
```
cathmer, line 29: I assume this code will be replaced by the code of PR #95 ?

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
```diff
@@ -0,0 +1,308 @@
...
+import java.util.ResourceBundle;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Draws two compared graphs above each other in a pane.
+ *
+ * @author Faris
+ */
+public class DrawComparedGraphs implements Initializable {
```
cathmer, line 31: I feel like there is a lot of logic in this class, while it is a GUI class. Should this logic not be moved somewhere else? I'm not sure where though, maybe @Pathemeous has an idea about this.

cathmer:
I reviewed most of the code. 

I didn't extensively review the DrawComparedGraphs class though, because it's a lot of code and it's not very easy to understand. Maybe this code should be split up into different classes/more methods to make it more readable. 

Overall the code looks good, there is some nice use of new Java features and the logic in the algorithms is very sound.

Don't forget testing your classes though.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/GraphOrdererThread.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Thread which can be used to order a graph.
+ *
+ * @author Faris
+ */
+public class GraphOrdererThread extends Thread {
+
+  private HashMap<Integer, GraphNodeOrder> orderedGraph;
+  private final OriginalGraph graph;
```
cbilstra, line 22: Maybe use GraphInterface if possible to make sure that also a BubbledGraph can be ordered?

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/RootLayoutController.java
```diff
@@ -102,24 +188,54 @@ private void initializeGraphIcon() {
...
+   * Thread which is used to get a graph of a subset of the genomes from a graph.
    */
-  public void setData(IPhylogeneticTreeNode root) {
-    assert treeManager == null;
-    treeManager = new TreeManager(treePane, root, zoomOutButton, selectionManager);
-    heatmapManager.initLeaves(treeManager.getCurrentLeaves());
-    treeManager.setOnLeavesChanged((Observable observable, Object arg) -> {
-      heatmapManager.setLeaves(treeManager.getCurrentLeaves());
-    });
+  private class SplitGraphsThread extends Thread {
```
cbilstra, line 201: Maybe also put this in a separate file?

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
cbilstra, line 31: There is much logic contained in this class but the logic has to do with drawing the graph. Thus, I think we can leave this code here. It could be an option to put some of it in a controller class though?

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/CompareSubgraphs.java
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    /**
+     * Get all of the nodes which are in both the top graph and the bottom graph.
+     *
+     * @param topGraph    the top graph.
+     * @param bottomGraph the bottom graph.
+     * @return the overlapping nodes.
+     */
+    private static HashMap<Integer, AbstractNode> calculateOverlappedNodes(OriginalGraph topGraph,
```
cathmer: Why is this method static?

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/CompareSubgraphs.java
FarisElghlan, line 114: I don't agree. This class is only used by the CompareSubgraphs class and is only used to execute a single method on a separate thead. If I wouldn't have multithreaded, this would simply have been 1 method, so in my opinion this class belongs within the CompareSubgraphs class.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/CompareSubgraphs.java
FarisElghlan, line 199: See comment above.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
FarisElghlan, line 29: Yes.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
FarisElghlan, line 31: Well this is a controller class. Fxml/css files are basically the pure visual representation components, the controller class of a fxml file decides what is actually shown in every part of the view (only the objects, usually not what they look like) and what actions are performed when the user interacts with objects in the view. There are however exceptions to this (for example if the shade of a color depends on a certain value).
The class basically only draws the tree, so this is pure GUI stuff. The only potential exception to this could be the `repositionOverlappingEdges` method, which shifts nodes up if inlinks are overlapping, but this also is a visual thing, so in my opinion belongs in this class.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/GraphOrdererThread.java
FarisElghlan, line 22: I think we should delay this till the model classes have been refactored. I think the bubbledgraph will dissapear, so we might as well remove the graphinterface and keep just the originalgraph.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/RootLayoutController.java
FarisElghlan, line 201: This class is only used by the RootLayoutController class and is only used to execute a single method call on a separate thead. If I wouldn't have multithreaded, this would simply have been 1 method call, so in my opinion this class belongs within the this class.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/CompareSubgraphs.java
FarisElghlan: First it was defined outside of this nested class, after refactoring I forgot to change it to non-static. Fixed.

wouterjsmit:
A tip for future commiting: try not to commit development/testing code such as timers. I see that you removed them in a later commit, but now they pollute the `Fixed node alignment bug` commit.

wouterjsmit:
Great, a bit messy history, especially with #95 sort of entangled in it, but not worth cleaning it up.

Code looks good and has been reviewed already, merging.

PR #95, opened 2016-05-16
Split Graphs

wouterjsmit:
Creates the Split Graph algorithm

@FarisElghlan Heavily improved the performance, after which I refactored the code and heavily unit tested it.

The only functionality that is not tested, is whether all fields of the nodes, such as `bases` and `alignment` are copied over. Note that the `inLinks`, `outLinks` and `genomes` **are** tested.

### DIFF
--- a/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
+++ b/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
@@ -6,8 +6,8 @@ import nl.tudelft.pl2016gr2.thirdparty.testing.utility.TestId;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.NoSuchElementException;
 
 /**
@@ -19,7 +19,7 @@ import java.util.NoSuchElementException;
  * <li>The nodes in the subgraph are a subset of the nodes in the original graph.</li>
  * <li>The nodes in the subgraph will have an identical ID as their counterparts in the original
  * graph.</li>
- * <li>The nodes will be newly instantiated, meaning that the subgraph datastructure is not backed
+ * <li>The nodes will be newly instantiated, meaning that the subgraph data structure is not backed
  * by the original graph.</li>
  * </ul>
  * </p>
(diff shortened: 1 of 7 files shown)


PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
```diff
@@ -0,0 +1,207 @@
...
+  /**
+   * Builds a subGraph of the <code>mainGraph</code> containing only the nodes that contain at
+   * least one of the specified <code>genomes</code>.
+   *
+   * @param genomes The genomes that should be included in the resulting graph
+   * @return An <code>OriginalGraph</code> that is a subGraph of <code>mainGraph</code>
+   * @throws NoSuchElementException Iff the genome collection is not a subset of the main graph's
+   *                                genomes
+   */
+  public OriginalGraph getSubGraph(Collection<String> genomes) {
```
FarisElghlan: I think subgraph is 1 word, so this should be called `getSubgraph`?

PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphsTest.java
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  private void setupIterator() {
+    nodeIDs = new ArrayList<>();
+    nodeIDs.add(0);
+    nodeIDs.add(1);
+    nodeIDs.add(3);
+
+    @SuppressWarnings("unchecked") // Necessary when mocking generic types
+        Iterator<Node> mockedIterator = (Iterator<Node>) mock(Iterator.class);
```
FarisElghlan: You can leave out the cast: `(Iterator<Node>)`

PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphsTest.java
```diff
@@ -0,0 +1,320 @@
...
+   * Verifies that the {@link SplitGraphs#getSubGraph(Collection)} method returns a non-empty
+   * graph.
+   */
+  @Test
+  public void testGetSubGraphReturnsNonEmptyGraph() {
+    ArrayList<String> genomeSub = new ArrayList<>(
+        Arrays.asList(new String[] {genomeSet.get(0)}));
+
+    OriginalGraph subGraph = defInstance.getSubGraph(genomeSub);
+    assertFalse(subGraph.getNodes().isEmpty());
```
FarisElghlan: You made the `getNodes()` method deprecated in this branch, while in this test file you use this method 11 times. I suggest you create a private static method in this test class which gets the field, using the AccessPrivate class:

``` java
  private static HashMap<Integer, Node> getNodes(OriginalGraph graph) {
    return AccessPrivate.getFieldValue("id", OriginalGraph.class, graph);
  }
```

FarisElghlan:
Looks great! Found only a few minor issues.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/OriginalGraph.java
```diff
@@ -80,6 +124,10 @@ public Node getRoot() {
     return nodes.get(lowestId);
   }
 
+  /**
+   * @deprecated Use {@link #iterator()} instead. This method exposes the hashmap.
+   * {@inheritDoc}
+   */
```
FarisElghlan: Iterators don't give any way to perform the action: nodes.contains(key) efficiently, so this would have to be implemented by this class if you want to remove this method.

cathmer:
The splitting algorithm looks good! Also looks like it has been extensively tested. This can be merged.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
```diff
@@ -0,0 +1,207 @@
...
+    if (!mainGraph.getGenoms().containsAll(genomes)) {
```
cbilstra: Would it be usefull to think about another option for containsAll? Time complexity of containsAll is O(n*m), which might be a bottleneck?

PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphsTest.java
cbilstra: I think there shouldn't be an indentation at line 65 (Iterator<Node> mocked....)

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
FarisElghlan: Nope. This action is performed only twice when comparing graphs (once for the top graph and once for the bottom graph), this won't be a bottleneck at all

PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphsTest.java
wouterjsmit: @FarisElghlan IntelliJ complained if I didn't cast it and some source on the Internet confirmed that when mocking generic classes, you must cast them, but I will have a look

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/SplitGraphs.java
wouterjsmit: We could make it an assert instead, this makes sure that it errors during development (when it fails), but is no longer checked when we build a release, leaving the check out of the final release and thus its performance.

However this depends on the question: Can we ensure that this will never ever happen in release code?

@cbilstra @FarisElghlan

wouterjsmit:
All feedback has been incorporated and the commits have been squashed. Merging this into dev.

PR #96, opened 2016-05-17
Feat/gui visualize base and genome quantity

### LINKED ISSUE #87: Edge thickness -> amount of genomes
Let the edge thickness depend on the amount of genomes which goes through them.

### LINKED ISSUE #88: Node size -> amount of bases
Let the size of the nodes depend on the amount of bases inside it.

FarisElghlan:
Edge thickness now depends on the amount of genomes going through them.
Node size now depends on the amount of bases in the node.

Fixes #87
Fixes #88


FarisElghlan:
Builds on #94, so wait with merging this till that PR is merged.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
```diff
@@ -225,26 +242,52 @@ private static void drawNode(Pane pane, HashMap<Integer, NodeCircle> circleMap,
...
+   * Calculate the radius of the node. The readius depends on the amount of bases inside the node.
+   *
+   * @param node the node.
+   * @return the radius.
+   */
+  private static double calculateNodeRadius(GraphNodeOrder node) {
+    int amountOfBases = node.getNode().getSequenceLength();
+    double radius;
+    if (amountOfBases > 1000) {
+      radius = Math.log(amountOfBases) * 4.0 - 17.0;
```
cathmer: These are magic numbers. Also, some explanation about the choice of number would be nice.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
```diff
@@ -225,26 +242,52 @@ private static void drawNode(Pane pane, HashMap<Integer, NodeCircle> circleMap,
...
+      if (radius > MAX_NODE_RADIUS) {
+        return MAX_NODE_RADIUS;
+      }
+    } else {
+      radius = Math.log(amountOfBases) * 0.7 + 5.1;
```
cathmer: Same here

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
```diff
@@ -256,18 +299,37 @@ private static void drawEdges(Pane pane, ArrayList<GraphNodeOrder> graph, int st
...
+   * Calculate the edge width.
+   *
+   * @param maxGenomes the total amount of genomes in the graph.
+   * @param from       the node from which the edge comes.
+   * @param to         the node to which the edge goes.
+   * @return the edge width.
+   */
+  public static double calculateEdgeWidth(int maxGenomes, AbstractNode from, AbstractNode to) {
+    int genomesOverEdge = from.getGenomesOverEdge(to);
+    double edgeWith = Math.log(100.0 * genomesOverEdge / maxGenomes) * 0.8;
```
cathmer: Again magic numbers

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/Node.java
```diff
@@ -55,12 +56,38 @@ public int hashCode() {
...
-    this.genomes = gs;
+  /**
+   * Get the amount of genomes which travel from this node over an edge to the given node.
+   *
+   * @param to the node to which the edge goes.
+   * @return the amount of genomes which travel over the given edge.
+   */
+  @Override
+  public int getGenomesOverEdge(AbstractNode to) {
+    assert getOutlinks().contains(to.getId());
```
cathmer, line 71: Assert's are considered bad practice. Instead, you could either return 0 when there is no outlink to the other node, or throw an exception. 
Also, to be 100% secure you should also check if the node where it's going to contains an inlink with this node.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/Node.java
```diff
@@ -55,12 +56,38 @@ public int hashCode() {
...
+
+  /**
+   * Set the genomes of this node.
+   *
+   * @param genomes the genomes.
+   */
+  public final void setGenomes(Collection<String> genomes) {
+    this.genomes = new HashMap<>();
+    for (String genome : genomes) {
+      this.genomes.put(genome, genome);
```
cathmer: If you use the same string for value and key, why not use a hashset?

cathmer:
Nice feature. I only reviewed the code of the last commit, other code is reviewed in other PR's. 

I added some small comments, overall it looks good. However, I do think the DrawComparedGraphs class is becoming too big and contains too much logic for a gui class. You should consider refactoring this class and splitting the logic into multiple classes. This would make it a lot more readable.

PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/FilterSnipsTest.java
```diff
@@ -27,7 +27,7 @@
   @Before
   public void setup() {
     for (int i = 1; i <= 9; i++) {
-      Node node = new Node(i, 1, null, 0);
+      Node node = new Node(i, 1, new ArrayList<>(), 0);
       if (i == 1 || i == 4 || i == 7) {
```
cbilstra, line 31: if (i - 1 mod 3 == 0) kan je doen, dan is het duidelijk dat het om elke 3 nodes vanaf 1 gaat en heb je geen magic numbers meer. Weet niet of checkstyle dat ook in de tests bekijkt

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/RootLayoutController.java
```diff
@@ -44,31 +62,99 @@
...
-    assert (controller == null);
-    controller = this;
+  @Override
+  public void initialize(URL location, ResourceBundle resources) {
     heatmapManager = new HeatmapManager(heatmapPane);
     initializeSelectionManager();
     initializeTreeIcon();
     initializeGraphIcon();
-    new DrawGraph().drawGraph(graphPane);
+    graph = new GfaReader("TB328.gfa").read();
```
cbilstra: We should change the string to a variable I guess? Eventually we want to select a file at stratup

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
```diff
@@ -0,0 +1,370 @@
...
+    this.bottomGraphOrder = bottomGraphOrder;
+    int highestTopLevel = topGraphOrder.get(topGraphOrder.size() - 1).getLevel();
+    int highestBottomLevel = bottomGraphOrder.get(bottomGraphOrder.size() - 1).getLevel();
+    if (highestTopLevel > highestBottomLevel) {
+      amountOfLevels = highestTopLevel;
+    } else {
+      amountOfLevels = highestBottomLevel;
+    }
+    scrollbar.setUnitIncrement(1.0 / amountOfLevels);
+    scrollbar.setBlockIncrement(10.0 / amountOfLevels);
```
cbilstra: Magic number?

PL2/PL2-launcher/src/main/java/nl/tudelft/pl2016gr2/launcher/Dnav.java
```diff
@@ -35,32 +34,27 @@
    */
   @Override
   public void start(Stage primaryStage) throws IOException {
-    FXMLLoader loader = new FXMLLoader();
-    loader.setLocation(getClass().getClassLoader().getResource("pages/RootLayout.fxml"));
-
-    Scene scene = new Scene(loader.load(), 1000, 800);
+    RootLayoutController rootLayout = RootLayoutController.loadView();
+    Scene scene = new Scene(rootLayout.getPane(), 1000, 800);
     primaryStage.setMinHeight(700);
```
cbilstra: Maybe use variables for these integers?

PL2/PL2-core/src/test/java/nl/tudelft/pl2016gr2/core/algorithms/FilterSnipsTest.java
FarisElghlan, line 31: Dat is code van casper, heb alleen deze test aangepast omdat die faalde door null.
`i % 3 == 1` lijkt me handiger

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
FarisElghlan: Well they actually really are magic numbers, just generated a few graphs to check which function kind of nicely maps the size (which can vary between 1 and ~300 000) to a pixel size, not really sure what to do with it, only thing I can think of is to add this description as comments to the method.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/Node.java
wouterjsmit, line 71: `assert`s are not bad practice, they are similar to `exception`s, except (punny) that an `exception` may occur during run time. This function is an interaction with the data model. If we (the developers) call the model wrongly, we might trigger this assertion. However, this should **never possibly be triggered** in release code. Asserts in this projects are rather vital because they are not checked during the release, so they do not hinder performance. They still can trigger during development (when you compile with assert statements turned on), to notify of erreoneous behaviour.

In general: The caller is responsible for making sure that the provided `AbstractNode to` actually is an `outLink`. If you do not, you crash the application. Asserts just make it easier to see that you did :smile:

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/RootLayoutController.java
wouterjsmit: @justinvdk is working on this AFAIK

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
```diff
@@ -112,8 +122,11 @@ public void drawGraphs(ArrayList<NodePosition> topGraph,
...
   }
 
+  /**
+   * Update the graph by redrawing it.
+   */
   private void updateGraph() {
-    if (topGraph == null || bottomGraph == null) {
+    if (topGraphOrder == null || bottomGraphOrder == null) {
       return;
     }
```
wouterjsmit, line 133: I must have missed this in earlier PRs, but what is the meaning of this exactly? Seems like some sort of error handling.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
```diff
@@ -154,37 +170,37 @@ private static void drawGraph(Pane pane, ArrayList<NodePosition> graph, int star
...
    * index by 20 to keep some margin at the left of the screen (so the edges are drawn correctly)
    *
-   * @param graph      the graph.
+   * @param graphOrder the graph node order.
    * @param startLevel the start level.
    * @return the start index.
    */
-  private static int calculateStartIndex(ArrayList<NodePosition> graph, int startLevel) {
+  private static int calculateStartIndex(ArrayList<NodePosition> graphOrder, int startLevel) {
     int actualStartLevel = startLevel - OFFSCREEN_DRAWN_LEVELS;
```
wouterjsmit, line 189: The comment says `20`, but the value of `OFFSCREEN_DRAWN_LEVELS` is `10`. Let the comment link to the constant with `{@link #OFFSCREEN_DRAWN_LEVELS}`

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
```diff
@@ -225,26 +242,52 @@ private static void drawNode(Pane pane, HashMap<Integer, NodeCircle> circleMap,
   }
 
   /**
+   * Calculate the radius of the node. The readius depends on the amount of bases inside the node.
```
wouterjsmit: `readius -> radius`

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/Node.java
FarisElghlan: Didn't know that a hashset should be used in these cases, thanks!

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
wouterjsmit: Simply put some comments around these magic numbers explaining what they do and why they're set to this value. Putting them in constants wouldn't actually help with understanding them, only move the `WTF?` moment to the constant definition rather than where the constant is used.

PL2/PL2-shared/src/test/java/nl/tudelft/pl2016gr2/model/NodeTest.java
```diff
@@ -60,7 +61,9 @@ public void basesTest() {
   
   @Test
   public void genomeTest() {
-    assertEquals("ref2.fasta", node.getGenomes().get(1));
+    Iterator<String> it = node.getGenomes().iterator();
+    it.next();
+    assertEquals("ref2.fasta", it.next());
   }
```
wouterjsmit, line 67: Big `Yeay` for API testing. Let's not break tests when I refactor the `model` package, so I don't have to spend hours trying to rewrite them

wouterjsmit:
@FarisElghlan I made a few minor comments, also to @cathmer's previous comments.

I particularly like that you did some work in fixing some tests (using iterators instead of gets, etc.). Will save me work later :). (although I do believe that `get()` is part of the `Collection<>` interface, so it is not really different or better, but the idea is there)

One thing I must comment, but have no fix for myself, is the amount of mindblowing calculations. I guess this is simply a GUI thing, but it would be useful to maybe add more line comments, or if possible, split up in small methods / chunks that are more comprehensible on their own.
However I understand that this is hard and I'm sure you've heard this a few times before now, so you'll have thought about it already.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/DrawComparedGraphs.java
FarisElghlan, line 133: If there isn't any graph being shown in the graph view, it shouldn't be updated.

FarisElghlan:
Updated code according to feedback.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/RootLayoutController.java
justinvdk: Correct, see #100. This PR has lots of conflicts I'd like resolve with 4 eyes as both changesets are pretty big and move code around.

PR #97, opened 2016-05-18
Feat/gui tree graph layout

### LINKED ISSUE #89: Tree to graph transition
Create a better transition fom the tree to the graph. The tree should still be visible when viewing the graph, so the user can see what is being compared in the graph.

### LINKED ISSUE #92: Phylogenetic tree edge length
Use the length of the edges when drawing the phylogenetic tree, so the evolutionary distance is derivable from them (or the edges must at least give an estimation of the evolutionary distance).

FarisElghlan:
The tree and graph are now drawn in the same view.
The tree is drawn from right to left instead of from left to right, because now the heatmap can be draw at the left side (having the heatmap between the tree and graph looks really ugly).
The tree now uses the length of the edges which is read from the newick file.

Fixes #89 
Fixes #92


FarisElghlan:
Please wait with merging till #96 has been reviewed and merged, as this branch builds on #96.

cathmer:
Nice addition. I don't have any comments on the code right now, but it's a little hard to review it because the number of added lines is so big, and spitting through the commits is also not so convenient when reviewing.

Maybe others can review this later when the other PR's have been merged into dev.

cbilstra:
I agree with @cathmer that it's difficult to see the difference with the other PR's. Will review this later?

FarisElghlan:
Update: merged other PRs into dev and dev into this branch.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/RootLayoutController.java
```diff
@@ -79,13 +62,35 @@
...
     graph = new GfaReader("TB10.gfa").read();
     mainGraphOrder = new GraphOrdererThread(graph);
     mainGraphOrder.start();
+    loadTree();
+  }
+
+  /**
+   * Load the data into the root layout.
+   *
+   * @param controller the controller of the root layout.
```
wouterjsmit, line 79: There are no parameters for `loadTree()`?

FarisElghlan, line 79: Yes, the file path is hardcoded because there is no file selection option yet. When this feature is implemented the `loadTree()` method can easily be adjusted though.

wouterjsmit:
Looks good, I found nothing extraordinary.

Issue #82, opened 2016-05-15
Time Scalability: Data parsing (and processing) in real time


wouterjsmit:
## Introduction

One of the main goals of this project is to achieve a data parsing (and processing) system that is scalable. Scalable can be viewed in terms of space complexity and time complexity.
This issue looks at **time complexity**.

This issue serves as a central point of discussion any development that is related to this goal. Any thoughts, progress, benchmarks, analytics, pull requests or issues can be posted / linked here.
## Grand Goal

Currently, scalable time complexity means that the system is able to load a 6000 genome sequence graph in real time. This is the highest benchmark that the client has provided us. However, achieving this benchmark is not sufficient for complete scalability. Complete scalability means that the system will be able to load future data sets in real time as well. These data sets are estimated to grow explosively, with an expectation of a total of 60,000 genomes next year. This is a 1000% increase per year.
## Definition of 'Real time'

Because genome data sets are inherently large, dealing with the data is no trivial point. While 'real time' is often defined as _instantenous to the user_, we must therefore make a slight adjustment to this definition.

We allow for the data to be pre-processed (parsed, loaded in) at the start of the application. This may take more than minutes (but not hours if possible). Generally speaking, this step may take _reasonable_ time, which can be roughly translated as _as fast as possible within our capabilities, budget and schedule_.

The point of pre-processing is that it only happens once: at the start of the workflow. During any further use of the application, all interaction should provide an _instanteneous_ response. The pre-processing may take longer, if that is necessary to achieve this.

wouterjsmit:
#81 creates a big leap in this goal.
@FarisElghlan also provided a benchmark and suggestions for future work regarding this problem in a [comment](https://github.com/ProgrammingLife2016/PL2-2016/pull/81#issuecomment-219294869).

Quoting him:

> Performance:
> First parsing the dataset of 10 genomes took around 2.5 seconds, now it takes around 100-120 milliseconds (20-25 times faster)
> First parsing the dataset of 328 genomes took around 3 minutes, now it takes around 1.5-1.7 seconds (105 times faster)
> 
> The algorithm has a time complexity of O(n) (lineair), so I expect the dataset of 6000 genomes to be loadable within 30 seconds (6000 / 328 \* 1.7 = 31 seconds).
> 
[...]

PR #100, opened 2016-05-19
Feat/open file dialog

justinvdk:
@FarisElghlan Seems like my branch can't merge into dev anymore. I can't quite see how I should merge as we both changed quite a lot in when/how the tree+graph get loaded into the controller of the main screen.

Maybe we should merge together, so we won't loose functionality somewhere.


justinvdk:
Furthermore a commit in this branch adds 'oracle-java8-installer' as apt-package to travis. This ensures a new version for java is used, and apparently I used some classes of JavaFX which were only added later in JDK8

FarisElghlan:
Is it ok if I review this tomorrow? I have been busy for the last 3 hours implementing feedback/merging my PRs and still have to write a lot of tests.

wouterjsmit:
We might not actually want to implement this at all, [see comment here](https://github.com/ProgrammingLife2016/PL2-2016/issues/82#issuecomment-220645937).

Would this be useful on its own? If so we obviously do not need to purge it.

FarisElghlan:
@Pathemeous 
This PR implements a way to open the tree/graph file via the UI instead of giving it as a hardcoded path in the RootLayoutController class. It doesn't have to do anything with performance/data storage (well, except that a preference file/last opened files list might have to be stored on disk).

FarisElghlan:
@justinvdk 
I tested it and it works well, the only problem I had was that the user is required to give a graph and tree path separately. I think there will be around two more files which have to be imported, so that's a lot of effort. This also causes a bit of a problem when the user opens the file browser: am I selecting the tree or the graph file? (at least I had this problem, when I selected one of the buttons and was searching for a file I forgot which one I pressed). 
I think we should use a directory browser (with which you can select a directory instead of a file) to select all of the files which are needed at once. We can then create a text field (like you did) for all of the files which are needed, when the user selects a directory , all of the needed files which are present in the directory are selected. If any of the files are missing, the user can select the file manually with a filebrowser (or change selected files if the directory for example contains multiple gfa files or an incorrect gfa file).
Like this you can import all of the files quickly if they are in the same directory, but still change some of the files if needed. (Selecting the directory should be very similar to the way geex lets the user select the files).

justinvdk:
> am I selecting the tree or the graph file? 

I agree, in code I do set a title property but it seems that this wasn't exactly what I expected. I assume you tested on Windows so this isn't only on osx.

To be clear: your suggestion means to let the user select a directory, and in this directory a search is done for all the needed files (by extension). So only when ambiguous or erroneous would the user select individual files? I think this is a great idea. I was wondering what would be a good approach as I agree that it is too tedious UX wise.

This also tackles the need for a third file when metadata is implemented.

todo:
- make known to the user what he/she is selecting in the (native) file dialog.
- process suggestion

FarisElghlan:
It looks like this on windows:
file browser
You can see the title, but it is overlooked most of the time.
Also found out two more things:
- The title of the file browser is incorrect when clicking the open file button of the graph (it says "Open tree File" for both the tree and graph browser, also I think File shouldn't be written with a capital, since tree isn't either).
- When you click the cross arrow in the top right corner, of the "Select files to load" window the application crashes (because the window is closed and in this case you probably assume the correct file has been selected). You could remove the close button by doing this: `stage.initStyle(StageStyle.UNDECORATED);`, but this way it will be impossible to close the application if you don't select a file, so it would probably be a better idea to either close the application if the close button is clicked (maybe with an additional warning window "do you really want to close the application?"), or check if the values in the text fields are correct when the window is closed (so the application doesn't crash). I think the last option is the best one.

justinvdk:
- That is a mistake on my side, thanks this creeped in later in the process as I'm sure I had it correct the first time around ;). What is weird is that the title DOES seem to work on windows.. On OSX I just see the generic title of the file browser.
- Thanks, that is useful

justinvdk:
Meh. Github doesn't allow PR's te be reopened so my branch deletion (prior to re-push) kinda broke this thread.

wouterjsmit:
Not sure what you mean? If it was local you lost it. If it was ever on GitHub you could probably restore it somehow.

wouterjsmit:
Please open new thread though :-)

PR #102, opened 2016-05-20
Write tests for the Pair and NodePosition classes

cbilstra:
I added two test classes, they contain unit tests and have 100% line coverage


FarisElghlan:
Nice tests. I don't really have any comments on your code (looks good).

PR #103, opened 2016-05-20
Test/shared model

justinvdk:
This PR adds some more tests for the shared module


cbilstra:
We are lucky that checkstyle does not check for magic numbers when looking at test classes  ;p. I assume it has 100% coverage?

FarisElghlan:
Nice tests. I don't really have any comments on your code (looks good).

PR #105, opened 2016-05-20
Tested the compare graph classes

FarisElghlan:
Added tests for the compare graphs classes of the core module.


cbilstra:
Tests look fine to me. @Pathemeous I think you can merge this?

PR #107, opened 2016-05-20
Fix adding genomes

justinvdk:
Fix a bug that occurred after the merge of two different PRs


wouterjsmit:
(Either of )you added this later, because this was not originally part of the split graph algo. Should have made a test to catch missing things like these :)

PR #108, opened 2016-05-20
Doc/update testing document

FarisElghlan:
Updated the testing document according to the changes made during sprint 4.


wouterjsmit:
PR only contains documentation: travis fails because it could not access the repository.

PR #111, opened 2016-05-21
Add 'fact' branch group for refactoring purposes

wouterjsmit:
'misc' would be too broad and other groups do not cover the definition of refactoring either.


wouterjsmit:
Merging since no resistance.

PR #121, opened 2016-05-22
Feat/gui drag tree nodes

### LINKED ISSUE #90: Compare arbitrary set of nodes
Make it possible to select an arbritrary set of nodes to compare by dragging and dropping nodes of the phylogenetic tree into the graph area.
Subtasks:
- [x] drag and drop tree nodes
- [x] clear old graph and add the dropped nodes
- [x] add the dropped nodes to the currently drawn graph
- [x] add the dropped nodes to a new graph (which is inserted below)
- [x] create a clear visual representation for the actions above and make sure everything is intuitive/looks nice

FarisElghlan:
Implemented drag n' drop for phylogenetic tree nodes to the graph area.

Fixes #90


wouterjsmit:
`+1,478 −727` :cry:

FarisElghlan:
A lot of those changes are actually because of class renames/moving methods (but still quite a lot of additions yeh)

wouterjsmit:
Will review soon. Midterms do come in the way. Is this branch needed by some other work? Aka, is there any need to branch off this PR or can future work be done from `dev`?

FarisElghlan:
There were some renames and a lot of lines have been changed, so I prefer if this is merged soon. To make reviewing easier I will create a list of classes which got important changes, all of the other classes mainly contain renames/moved methods and sometimes a little bit different implementation which does exactly the same thing.
List of classes to check:

These classes now have a persistent tree structure (i.e. the tree is not recreated every time and each node remembers its parent/children). Before this was managed by the tree parser dependency classes (which we didn't write ourselves)
- IPhylogeneticTreeNode
- IPhylogeneticTreeRoot
- PhylogeneticTreeNode
- PhylogeneticTreeRoot

This class can now draw different kinds of graph (i.e. a single graph, or both graphs) and add/remove genomes from any of the graphs very easily. This is done using an observable hashset which contains the names of the drawn genomes. This observable hashset is initialized in the selection manager (and retrieved in the setSelectionManager method) and is listened to by the TreeManager - to update the phylogenetic tree classes (so the selected nodes in the phylogenetic tree match the drawn genomes in the graph) and the TreeNodeCircle listens to the phylogenetic tree for changes of the booleanproperty (if the node is drawn in the top / bottom graph) and changes its colors accordingly.
- DrawComparedGraphs
- TreeManager
- TreeNodeCircle

The other classes are not that important to look at.

FarisElghlan:
If this description makes reviewing easier/quicker I could also do it for other PRs by the way.

wouterjsmit:
@FarisElghlan It definitely does make it easier. Such a description (maybe a little shorter) could we well fit for the PR message.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/model/IPhylogeneticTreeRoot.java
```diff
@@ -0,0 +1,27 @@
...
+   * Notify the tree root that a genome has been added to (isDrawn = true), or removed from (isDrawn
+   * = false) the top graph.
+   *
+   * @param genome  the genome which has been added or removed.
+   * @param isAdded if the given genome is added or removed.
+   */
+  void setDrawnInTop(String genome, boolean isAdded);
+
+  /**
+   * Notify the tree root that a genome has been added to (isDrawn = true), or removed from (isDrawn
```
wouterjsmit, line 20: `isDrawn ->`isAdded` See [line 11](https://github.com/ProgrammingLife2016/PL2-2016/pull/121/files#diff-47fe253a6cf9722928e62d565e8a80ffR11) as well.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/model/PhylogeneticTreeNode.java
```diff
@@ -1,109 +1,240 @@
...
- * This class is an adapter class which maps the methods of the
- * net.sourceforge.olduvai.treejuxtaposer.drawer.TreeNode class to the interface which is needed by
- * our application.
+ * This class is a storage class for a phylogenetic tree node. It retrieves and stores the needed
+ * values of the newick parser {@link TreeNode}.
  *
  * @author Faris
  */
-public class PhylogeneticTreeNode implements IPhylogeneticTreeNode {
+public class PhylogeneticTreeNode implements IPhylogeneticTreeNode, Iterable<PhylogeneticTreeNode> {
```
wouterjsmit, line 17: Is the `Iterable<>` interface not a property of the `TreeNode` _interface_? Unless it is only specific to this implementation, you should do `IPhylogeneticTreeNode extends Iterable<IPhylogeneticTreeNode>` instead, to enforce all implementations to be iterable. Note that the iterable now is over the _interface_ rather than the implementation (mind the `I`).

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/model/PhylogeneticTreeNode.java
```diff
@@ -1,109 +1,240 @@
...
+    this.weight = node.weight;
+    this.parent = parent;
+
+    if (node.numberChildren() == 2) {
+      children = new PhylogeneticTreeNode[2];
+      children[0] = new PhylogeneticTreeNode(node.getChild(0), this);
+      children[1] = new PhylogeneticTreeNode(node.getChild(1), this);
+    } else {
+      children = null;
+    }
```
wouterjsmit, line 51: Document why this if-else clause is how it is, Does the `TreeNode` interface only allow either `2` or `0` children or something?

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/model/PhylogeneticTreeNode.java
```diff
@@ -1,109 +1,240 @@
...
  *
  * @author Faris
  */
-public class PhylogeneticTreeNode implements IPhylogeneticTreeNode {
+public class PhylogeneticTreeNode implements IPhylogeneticTreeNode, Iterable<PhylogeneticTreeNode> {
+
+  private final String label;
+  private final float weight;
+  private final PhylogeneticTreeNode[] children;
+  private final PhylogeneticTreeNode parent;
```
wouterjsmit, line 22: Does the node need to know about its `parent`? It doesn't seem used anywhere except for the getter/constructor.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/tree/TreeManager.java
```diff
@@ -232,14 +235,12 @@ private void initializeGraphSizeListeners() {
    * @param root the root of the part of the tree which should be shown.
    */
   @TestId(id = "setRoot")
-  private void setRoot(IPhylogeneticTreeNode root) {
+  public void setRoot(IPhylogeneticTreeNode root) {
```
wouterjsmit: Probably want to rename this to `setCurrentRoot`. I was wondering why in the constructor you do `rootNode =` and then call this method.

wouterjsmit:
Made some comments. Overall the code looks good to me.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/model/PhylogeneticTreeNode.java
FarisElghlan, line 17: It is only used by the root node to parse over all of the leaf nodes and add them to a (genomename -> node) mapping. It is not used by any other classes/for any other purpose.
I used an iterator because it is well known (so people know what it does). The only thing which might be a little bit different from the usual iterator is that it only parses over the leaf nodes (so non-leaf nodes of the tree are skipped).

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/model/PhylogeneticTreeNode.java
FarisElghlan, line 51: A newick file can have multiple children per node, but a phylogenetic tree can't. Since the class is called `PhylogeneticTreeNode` I think this is clear enough already. I don't think it's really needed, but I will add two line comments.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/model/PhylogeneticTreeNode.java
FarisElghlan, line 22: It is also used in the `hasParent()` method. The UI stores the current root node (which is the root of the part of the tree which is currently shown on the screen), so to zoom out the parent is needed.
The `setDrawnInBottom` and `setDrawnInTop` also use the parent: to propagate the value of drawnInTop and drawnInBottom to the parent class whenever it changes (ex. when both children of a node are drawn in top, the node must be marked as drawn in top). This could be done with listeners, but I don't really see an advantage of listeners for this case.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/tree/TreeManager.java
FarisElghlan: The real problem with this line is that it is public and has a TestId though. I made it public because there was an exception when calling the method from a test and if you use a testid to call the method the real exception is not shown (a downside to reflection I guess), but I forgot to make it private again.
Good job for somehow finding it I guess ;)

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/model/PhylogeneticTreeNode.java
FarisElghlan, line 22: Edit: the UI actually also stores the original root, but finding the parent of a node in the tree by starting at the root requires `O(n)` time, since in the worst case you will have to browse through all nodes to find the current node and then return its parent. Like this it is `O(1)`.

FarisElghlan:
Can be merged.

wouterjsmit:
Cannot fast-forward it though. Could you resolve the conflicts?

PR #122, opened 2016-05-22
Feat/metadata reader

justinvdk:


PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/AnnotationReader.java
```diff
@@ -0,0 +1,211 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@SuppressWarnings("checkstyle:AbbreviationAsWordInName")
+public class AnnotationReader {
```
FarisElghlan, line 19: Could you add class javadoc which describe what the class does (and preferably also what `@SuppressWarnings("checkstyle:AbbreviationAsWordInName")` does and why it is needed).

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/AnnotationReader.java
```diff
@@ -0,0 +1,211 @@
...
+  private static final Logger logger = Logger.getLogger(AnnotationReader.class.getName());
+
+  private final File file;
+
+  public AnnotationReader(File file) {
+    this.file = file;
+  }
+
+  @SuppressWarnings("checkstyle:MethodLength")
+  private Annotation readRow(Row row) {
```
FarisElghlan, line 46: This is a long method. Add javadoc to describe what it does and why the annotation is needed.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/AnnotationReader.java
```diff
@@ -0,0 +1,211 @@
...
+
+  private void parseSpecimenID(Annotation annotation, Cell cell) {
+    annotation.specimenID = cell.getStringCellValue();
+  }
+
+  private void parseAge(Annotation annotation, Cell cell) {
+    if (cell.getCellType() == Cell.CELL_TYPE_NUMERIC) {
+      annotation.age = new Double(cell.getNumericCellValue()).intValue();
+    } else {
+      annotation.age = null;
```
FarisElghlan: This value is null by default, remove the else clause.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/AnnotationReader.java
```diff
@@ -0,0 +1,211 @@
...
+    return annotation;
+  }
+
+  private void parseSpecimenID(Annotation annotation, Cell cell) {
+    annotation.specimenID = cell.getStringCellValue();
+  }
+
+  private void parseAge(Annotation annotation, Cell cell) {
+    if (cell.getCellType() == Cell.CELL_TYPE_NUMERIC) {
+      annotation.age = new Double(cell.getNumericCellValue()).intValue();
```
FarisElghlan: Is there any reason for not writing `(int)cell.getNumericCellValue()`?
This seems to do exactly the same, since the implementating of `intValue()` looks like this:

``` java
public int intValue() {
    return (int)value;
}
```

justinvdk: No there wasn't, overlooked. Thanks!

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/AnnotationReader.java
justinvdk: Removed.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/AnnotationReader.java
justinvdk, line 46: I've added some extra javadoc.

PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/AnnotationReader.java
justinvdk, line 19: Initially this was added because a lot of column names in the sheet would cause warnings and I tried to keep the naming the same as much as possible. I've now checked again and only several methods/fields would now need this warning without changing too many names. I've added the warning to the specific cases now instead of the whole class.

I think with the leftover cases it's quite obvious for the developer why the warning is needed now as the names that cause the warning are directly below it.

justinvdk:
I've committed several changes.
@Pathemeous I assume you'd love to have a clean history, so tell me when (if?) you want to merge, then I'll cleanup a bit. Would you like to have everything squased or pom.xml / checkstyle.xml changes in a different commit ?

wouterjsmit:
@justinvdk Please do.
- [ ] Please split your fixup (feedback/javadoc) commits and amend each chunk to the correct original commit (javadoc for a class in the commit that originally implemented the class).
- [ ] 1 commit per project management related task (1 commit per dependency addition, 1 commit per checkstyle rule amendment, etc.)
- [ ] 1 commit per conceptual implementation. So 1 commit of actual business code in this case (`Implement Annotation Parser` or the likes)

Notes for this commit: 
- [ ] first line in message reading `Fixes: <url of issue>` The URL will be prettified to an issue number on GitHub.com and allows readers in a git log to actually copy the url and access your issue that way. A further message after another blank line if so desired (advisable is a 2-sentence summary of the issue).
- [ ] For sake of completeness: any commit message line that is not the title line must be wrapped at 72 characters (pro tip: set up your `~/.vimrc` to do this automatically by adding the following line:
  `autocmd Filetype gitcommit setlocal spell textwidth=72`)

Template:

```
<commit title, max 50 char>
 <blank line>
Fixes: <url to issue>
 <blank line>
Short summary of wtf you did
```

FarisElghlan:
Or we could just merge this so other people can use it for their tasks? Spending that much time to make commits look prettier seems like a waste of time to me (especially for a "top priority" task).

wouterjsmit:
It only looks like a lot of time because I wrote it down elaborately, but I'm just gonna squash as I do agree it takes too long before Justin has done it and then I have merged it, could've taken less than 5 minutes though

Issue #83, opened 2016-05-15
Model Classes must be refactored


wouterjsmit:
The model classes leave a lot of room for (space complexity) optimization.

The classes are defined by wrong inheritance structures, violating the Liskov principle where some API should be provided by parents rather than children.
The API of some classes does not hide the class implementation (for instance exposing the inner datastructure with `OriginalGraph.getNodes()`.

Doing this is necessary for the code quality, but also to improve the space complexity, which is probably our only remaining problem in terms of data scalability.

FarisElghlan:
Some of the model classes currently have low test coverage. Could you create an issue about this when you're done with the refactoring?

wouterjsmit:
I will add tests wherever possible.

wouterjsmit:
Since I am sick I will most likely not be able to finish this before 12:30 tomorrow. I will make sure to do it next iteration (which is also the version that we will deliver to SIG for software quality feedback).

wouterjsmit:
Closed by #124

PR #124, opened 2016-05-23
Refactor model classes

wouterjsmit:
Refactors the model classes into better design patterns and generally better maintainable inheritance hierarchies.

This PR will not compile by itself. See #125 for the changes required to use the new model classes in place of the old. I separated them so that the actual changes (this PR) would be separated from the hundreds of lines that had to be edited to achieve the port to this model structure.

Tip: Use the split diff view to see the complete new classes, since most files have been almost completely rewritten. The deletions that pollute the originally existing files can be confusing otherwise.

Fixes: #83


PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/HashGraph.java
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * Constructs a <code>HashGraph</code> with the specified elements and genomes.
+   *
+   * @param nodes   The nodes of the graph
+   * @param genomes The genomes that are represented in the graph
+   */
+  public HashGraph(Map<Integer, ? extends GraphNode> nodes, Collection<String> genomes) {
+    this.nodes = new HashMap<>(nodes);
+    this.genomes = new HashSet<>(genomes);
+    this.rootNodes = new HashSet<>(parseRootNodes());
```
FarisElghlan: This seems like unnecessary overhead, just let the `parseRootNodes()` method return a `HashSet`.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/HashGraph.java
```diff
@@ -0,0 +1,156 @@
...
+  @Override
+  public Collection<String> getGenomes() {
+    return genomes;
+  }
+
+  @Override
+  public void addGenome(String genome) {
+    assert !genomes.contains(genome) : "Adding already existing genome to the graph.";
+
+    genomes.add(genome);
```
FarisElghlan: Remove the whitespace (you don't have one here in your other methods)

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/StringSequenceNode.java
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * A simple implementation of <code>Node</code> that represents the DNA sequence as String.
+ *
+ * @author Wouter Smit
+ */
+public class StringSequenceNode extends AbstractNode {
+
+  private String sequence;
+  private HashSet<String> genomes;
```
FarisElghlan, line 15: We could maybe create a `ArrayList<Integer, String>` and `HashMap<String,Integer>` in the graph, which store `<id, genome name>` (where the arraylist index is the id) and `<genome name, id>` pairs and store only the ids of the genomes in the nodes. This will reduce storage and speed up the comparison of genomes a lot (comparing numbers is a lot faster than comparing strings).
Then we could also give the phylogenetic tree a reference to these hashmaps and use the same IDs there.

I tested reading a graph and checking how many characters were actually used to store the genome names in all of the graph nodes. It was actually a lot higher than I expected:
for the graph of 328 nodes there are 62 782 632 characters used to store the genome names in the nodes of the tree!
So that's 62 782 632 \* 2 = 120MB!

After this I have tried to determine the total amount of memory used by the graph by running this code:

``` java
GfaReader reader = new GfaReader("TB328.gfa");
System.gc();
System.out.println("before = " + (Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory()));
OriginalGraph graph = reader.read();
System.gc();
System.out.println("after = " + (Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory()));
```

resulting in:

```
before = 2726888
after = 583088568
```

Se around 553MB was used by the graph.

> the results of multiple runs were pretty consistens, within ~200kb

Then I tried exactly the same code, but I commented out the `setGenomes()` call in the parser (so everything was the same, except that the genomes weren't set for the nodes). This resulted in:

```
before = 2726816
after = 156414328
```

So around 147MB was used by the graph.

This is a reduction of 73.5%!!
The genomes names are 17 characters (34 bytes) long, so if we use integers (4 bytes) as IDs instead of the actual strings, we should be able to save around `1 - 4 / 34 = 88.2%` of the space which is currently used by the genome names, resulting in a reduction of `73.5% * 88.2% = 64.9%` of the memory used by the graph.

Since the references to the string are coppied when new graphs are created (the strings are not duplicated) this will result in a bit lower memory reduction when the application is actually doing stuff (splitting/comparing graphs), but even if this would reduce the memory by only 10% (probably still around 30-40% though) this would be worth it in my opinion.

> someone could make an issue out of this

FarisElghlan, line 15: Ok I got a better estimate for the size of the original graph if the genome names are replaced by ids, I simply changed every genome addition into an addition of a string of 2 characters (2 characters is the same size as an integer), resulting in:

```
before = 2750088
after = 174954064
```

So 164MB

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/StringSequenceNode.java
```diff
@@ -0,0 +1,165 @@
...
+  private HashSet<Integer> inEdges;
+  private HashSet<Integer> outEdges;
```
FarisElghlan: Currently the in and out edges are only used in loops (contains isn't called on them, besides assert statements). I also don't really see any reason to call contains on a set of edges, so we could implement this as an `ArrayList` and use the `trimToSize()` method when all of the edges have been added. This will ensure that the `ArrayList` doesn't use more space than is required (it removes the array entries which aren't used).

The `HashSet` uses: `32 * SIZE + 4 * CAPACITY` bytes (where `SIZE` is the amount of objects stored in the `HashSet` and `CAPACITY` is the total capacity of the `HashSet`), see: http://java-performance.info/memory-consumption-of-java-data-types-2/.
If you use an `ArrayList` with a call to `trimToSize()` instead, the amount of memory needed will be: `SIZE` bytes.

There are 131 777 links and 94 258 nodes in the graph of 328 genomes. Assuming every node has fewer than 9 in links and fewer than 9 out links, this would mean that every hashset uses the default capacity of 16.
With these assumptions, the memory usage for the hashset implementation will be:
`(32 * 131 777 + 4 * 16 * 94 258) * 2 = 20MB`
With the arraylist implementation this would be only:
`(4 * 131 777) * 2 = 1MB`

There isn't really any downside to using an `ArrayList`, except that the `trimToSize()` method must be called after all of the in and out links have been added.

20MB might not sound like much, but this will probably be over 20% of the graph size after all of the other memory improvements have been applied and the memory reduction will apply to all of the subgraphs which are made. If we load the graph of 6000 genomes (or more) it will of course save a lot more than 20MB for each graph aswell.

FarisElghlan:
Found a few style errors and a few ways to considerably improve the memory usage. The rest of the code looks really good!

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/StringSequenceNode.java
wouterjsmit: `contains()` is actuallly called quite a few times in the code to inspect the neighbours of a node. However, if the assumptation that every node has only few edges holds (and I do think it does), the lesser performance of `contains()` in a non-`HashTable` is not really a downside. I will implement this.

wouterjsmit:
@FarisElghlan I added a method `setInEdges()` and `setOutEdges()` which replaces the whole arraylist at once and calls `.trimToSize()` afterwards.
This way, the caller does not need to call the trimming function.

I left the `add` and `remove` methods, should it ever be needed to change those lists.

A slightly more philosophic question is whether those lists should be modifiable lists at all. This is another question for another PR though (the implementation could always be changed anyway, with the decoupling of the interfaces).

I did not yet implement the Genome (integer) references. I will do this later, as I'm afraid it might take more time to do nicely and this PR should really be merged soon. I made an issue here: #127. I (or someone else) could get to this later this week.

If there are no further comments, I will merge this.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/HashGraph.java
```diff
@@ -0,0 +1,154 @@
...
+   * @return all root nodes.
+   */
+  private HashSet<Integer> parseRootNodes() {
+    Collection<Integer> rootNodes = new ArrayList<>();
+    nodes.forEach((Integer elemId, GraphNode elem) -> {
+      if (elem.isRoot()) {
+        rootNodes.add(elemId);
+      }
+    });
+    return rootNodes;
```
FarisElghlan, line 70: The return type of this method is `HashSet<Integer>` while it actually returns `Collection<Integer>`

FarisElghlan, line 70: You should probably change the return type to `Collection<Integer>` and change the `rootNodes` variable to `Collection<Integer>` and all of the instantiations of `rootNodes` to `= new ArrayList<>()`

FarisElghlan:
Tested the new memory usage (by locally merging this branch into the `feat/implement-new-model-classes branch`) and my prediction was correct, exactly -20MB (+- 1MB) compared to before :)

FarisElghlan:
> I left the add and remove methods, should it ever be needed to change those lists.
> 
> A slightly more philosophic question is whether those lists should be modifiable lists at all. This is another question for another PR though (the implementation could always be changed anyway, with the decoupling of the interfaces).

@Pathemeous 
Currently only the copy method uses one of the constructors which doesn't set the in/out links. I am not sure how the copy method will be used, but maybe a setallinlink and setalloutlinks method could be implemented, or you can add a trimToSize method to the Node interface.
The add and remove methods are pretty usefull for testing purposes (easier to add a few edges like this than to initialize a list and set its elements).

wouterjsmit:
> Currently only the copy method uses one of the constructors which doesn't set the in/out links. I am not sure how the copy method will be used, but maybe a setallinlink and setalloutlinks method could be implemented, or you can add a trimToSize method to the Node interface.

`Copy()` must be implemented by the programmer who creates the implementing class. This means that he must also define the data structure, which means that he will know himself how to call it (whether to trim or not).

A `trimToSize()` method in the `Node` interface (or `GraphNode`) would be overkill and mainly shows that the lists in the class are too exposed.

> The add and remove methods are pretty usefull for testing purposes (easier to add a few edges like this than to initialize a list and set its elements).

Wouldn't it be better to add and remove lists using the `TestAnnotation` on the private fields (the `inEdges` and `outEdges`)? This way, you can call `add()` and `remove()` on the lists directly.

In production code there should just be no need for iteratively adding elements. I will deprecate those methods for that reason.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/HashGraph.java
wouterjsmit, line 70: I edited the `rootNodes` list to be an `ArrayList<>` rather than a `HashSet<>`. The helper function returns a new `ArrayList<>` which can directly be used as rootNode list. The `rootNodes` field is of type `ArrayList<>`, because it is pointless to hide the concrete  implementation from the class itself (by storing it as `Collection<>`). The `public` methods do return `Collection<>`s though.

Edit: not sure why the diff doesn't show as outdated though.

FarisElghlan:
Oh copy is actually used in splitgraphs, overlooked it.
Currently the code looks like this in splitgraphs:

``` java
GraphNode newNode = original.copy();
pruneGenomes(original, genomeSet).forEach(newNode::addGenome);
pruneInLinks(original, nodeSet).forEach(newNode::addInEdge);
pruneOutLinks(original, nodeSet).forEach(newNode::addOutEdge);
```

The only way to create a node with a trimmed arraylist for the subgraph (after you've removed the add/remove methods) would be to do this:

``` java
GraphNode newNode = original.copy();
newNode.getOutEdges().addAll(pruneOutLinks(original, nodeSet));
newNode.getInEdges().addAll(pruneInLinks(original, nodeSet));
((ArrayList)newNode.getOutEdges()).trimToSize();
((ArrayList)newNode.getInEdges()).trimToSize();
```

Also I don't really understand what you mean with:
"Copy() must be implemented by the programmer who creates the implementing class."
Since this method is already implemented in `StringSequenceNode`

And I don't agree that `trimToSize()` would expose the underlying storage type. All of the Collection classes which use an array which grows exponentially have a lot of empty entries. Hashmap doesn't offer a trim method though, since it is kind of hard to trim, removing all empty entries would result in the contains method to run in `O(n)` if the object is not in the map. But the fact that collections can have empty entries is still well known and not specific to a single implementation, so removing those entries isn't really exposing the type (if the implementation doesn't allow the data to be trimmed, it will simply do nothing).

wouterjsmit:
Yes, that should be refactored to simple use setEdges (which wasn't
previously available).

What I mean is:
When addEdges() or removeEdges() is not used, the data will never be
'untrimmed' (because the NodeGraph class encapsulates the maintenance of
the ArrayList that it uses). I deprecated those methods to leave an
in-between. If there seems legitimate use of them(in non tests), I will add
the trimming functionality in their implementations and remove the
deprecation.

The problem with those methods is that if you use them, you are either
hurting performance a lot (after each call, the addEdge method will call
trimToSize itself) or it will hurt memory a lot (trim is not called).
Instead of providing the ability to manually trim them (which as you show
is still possible), the setEdges methods provide an abstraction that deal
with both the adding and trimming afterwards.

Am on phone, apologies for bad formatting.

FarisElghlan:
Oh ok, I am working on the other branch, so `setOutEdges` and `setInEdges` weren't created yet. What I was basically saying was to add these methods, but you already did :+1:

PR #131, opened 2016-05-25
Replace Node type by GraphNode in SplitGraphs

wouterjsmit:
The old type of nodes in the graph were 'Node', but with the new model
interfaces, the type is 'GraphNode' instead. The tests should use this.
With this change, the graph tests work for any type of GraphNode.


FarisElghlan:
Looks good.

PR #132, opened 2016-05-25
Created a sequence class which stores the bases efficiently

### LINKED ISSUE #128: Sequences stored in GraphNode consume a lot of memory
There could be ways to improve this.
Thoughts are: Do not store the sequences at all or improve the representation to use less bits per base.
- Currently: String where each character is 1 base. (16 bits per character)

Points of thought:
- 5 options, needing 3 bits (with 3^8 - 5 = 3 dead states).
- When mapping the bits into an integer datatype, we still need an array of integers to allow for infinite length sequences. The array brings some overhead and should be minimized.
- The dead states can be used when doing bit shifts to mark the end of the sequence (let 000 mark the end).

FarisElghlan:
Fixes #128

performance of the 328 graph:
from: `580MB` and `3.5 seconds` 
to: `505MB` and `2.6 seconds`

Unexpected speed improvement aswell  :tada:

### DIFF
--- a/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
+++ b/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
@@ -1,0 +1,170 @@
+package nl.tudelft.pl2016gr2.model;
+
+import nl.tudelft.pl2016gr2.thirdparty.testing.utility.TestId;
+
+/**
+ * This is a container class for the bases of a genome. It efficiently encodes the ATCG and N
+ * (unknown) bases as bits in an array of integers, so fewer memory is needed than when using a
+ * simple String.
+ *
+ * @author Faris
+ */
+public class BaseSequence {
+
+  private static final int BITS_PER_BASE = 3;
+  private static final int BITS_PER_INT = Integer.BYTES * 8;
+
+  private static final int END_OF_BASES = 0b000;
+  private static final int BASE_A = 0b001;
+  private static final int BASE_T = 0b010;
+  private static final int BASE_C = 0b011;
+  private static final int BASE_G = 0b100;
+  private static final int BASE_UNKNOWN = 0b101;
+
+  @TestId(id = "bases")
+  private final int[] bases;
+
+  /**
+   * Create a base sequence.
+   *
+   * @param bases the sequence of bases.
+   */
+  public BaseSequence(String bases) {
+    this(bases.toCharArray(), 0, bases.length());
+  }
+
+  /**
+   * Create a base sequence.
+   *
+   * @param bases      an array containing the sequence of bases.
+   * @param startIndex the index in the array where the bases start.
+   * @param endIndex   the index in the array where the bases end.
+   */
+  public BaseSequence(char[] bases, int startIndex, int endIndex) {
+    int amountOfBases = endIndex - startIndex;
+    int bitsNeeded = amountOfBases * BITS_PER_BASE;
+    int intsNeeded = (int) Math.ceil(bitsNeeded / (double) BITS_PER_INT);
+    this.bases = new int[intsNeeded];
+    setBases(bases, startIndex, endIndex);
+  }
+
+  /**
+   * Get the base sequence as a string.
+   *
+   * @return the base sequence as a string.
+   */
+  public String getBaseSequence() {
+    StringBuilder sb = new StringBuilder();
+    for (int i = 0; (i * BITS_PER_BASE) / BITS_PER_INT < bases.length; i++) {
+      int encodedBase = getBase((i * BITS_PER_BASE) / BITS_PER_INT,
+          (i * BITS_PER_BASE) % BITS_PER_INT);
+      if (encodedBase == END_OF_BASES) {
+        break;
+      } else {
+        sb.append(decode(encodedBase));
+      }
+    }
+    return sb.toString();
+  }
+
+  /**
+   * Get a single encoded base.
+   *
+   * @param arrayIndex the index in the array of the base.
+   * @param bitIndex   the index of the starting bit of the base in the array.
+   * @return the read encoded base.
+   */
+  private int getBase(int arrayIndex, int bitIndex) {
+    if (bitIndex <= BITS_PER_INT - BITS_PER_BASE) {
+      return (bases[arrayIndex] >> bitIndex) & 0b111;
+    } else if (arrayIndex + 1 < bases.length) {
+      int bits = (bases[arrayIndex] >>> bitIndex) & 0b111;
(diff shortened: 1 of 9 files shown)


PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
```diff
@@ -0,0 +1,170 @@
+package nl.tudelft.pl2016gr2.model;
+
+import nl.tudelft.pl2016gr2.thirdparty.testing.utility.TestId;
+
+/**
+ * This is a container class for the bases of a genome. It efficiently encodes the ATCG and N
```
wouterjsmit: `This is a container class` comes over as if it is a data class. All classes are 'containers' for some sort of abstraction. Consider changing this to something along the lines of "Represents a DNA sequence."

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
```diff
@@ -0,0 +1,170 @@
...
+ * (unknown) bases as bits in an array of integers, so fewer memory is needed than when using a
+ * simple String.
+ *
```
wouterjsmit, line 10: Mention that the stored bases are immutable.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
```diff
@@ -0,0 +1,170 @@
...
+   * Create a base sequence.
+   *
+   * @param bases the sequence of bases.
+   */
+  public BaseSequence(String bases) {
+    this(bases.toCharArray(), 0, bases.length());
+  }
+
+  /**
+   * Create a base sequence.
```
wouterjsmit: I had to look at the parser to understand the parameters of the constructor. Please clarify that you create a base sequence _based on the graph file char array and its indices_.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
```diff
@@ -0,0 +1,170 @@
...
+
+  @TestId(id = "bases")
+  private final int[] bases;
+
+  /**
+   * Create a base sequence.
+   *
+   * @param bases the sequence of bases.
+   */
+  public BaseSequence(String bases) {
```
wouterjsmit, line 34: For this constructor you could specify what the String should look like (`ACTGN` only)

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+    return sb.toString();
+  }
+
+  /**
+   * Get a single encoded base.
+   *
+   * @param arrayIndex the index in the array of the base.
+   * @param bitIndex   the index of the starting bit of the base in the array.
+   * @return the read encoded base.
```
wouterjsmit, line 80: There is no mention of the return of `END_OF_BASES` when the index dit not yield a base in the array.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
FarisElghlan: Well it works for any array of characters which contains only ATCGN's, but it was indeed mainly made for the parser.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
```diff
@@ -0,0 +1,170 @@
...
+   */
+  private int getBase(int arrayIndex, int bitIndex) {
+    if (bitIndex <= BITS_PER_INT - BITS_PER_BASE) {
+      return (bases[arrayIndex] >> bitIndex) & 0b111;
+    } else if (arrayIndex + 1 < bases.length) {
+      int bits = (bases[arrayIndex] >>> bitIndex) & 0b111;
```
wouterjsmit, line 86: This shift operation uses the unsigned variant, but you seem to prefer the signed variant in the rest of the code. AFAIK they both have the same performance. Is there a reason for using two different operations? Could you not do something like create an unsigned integer and simply use `>>` everywhere?

wouterjsmit:
Great work! Beautiful code, including the tests. I made really minor comments, mostly about javadoc, since there is not really much to comment on.

wouterjsmit:
One thing I forgot: the `StringSequenceNode` should probably be renamed to something else, since it doesn't use strings anymore (it does use them at the API level, but that is defined in the `Node` interface, so if anything, that interface should contain the word `String`)

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
FarisElghlan, line 86: java doesn't even have unsigned integers :( (otherwise I would have used them of course). The >> operator shifts 0 bits in if a number is positive (i.e. starts with a 0 bit) and shifts 1 bits in if a number is negative (i.e. starts with a 1 bit). The >>> operator always shifts in 0 bits. This behavior isn't actually needed here, but it partly fixed a different bug and I didn't change it again because it works and I didn't try to change it back after fixing the actual bug.

wouterjsmit, line 86: I had a guess that this would go wrong for `>>` because of shifting in `1` bits. just to be sure, could you test it to see if it works with the signed shift? If so, it would be more consistent. Otherwise, obviously the unsigned shift is required and that be it.

FarisElghlan: Will change it back for clarity reasons.

cbilstra:
Had a quick look on it and looks good!

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/BaseSequence.java
FarisElghlan, line 86: Woops, it actually did matter for the "end of base" bits. In this case the bits "111" could be shifted in, instead of "000" which marks the "end of base".

wouterjsmit, line 86: In that case we should obviously leave it :).

PR #138, opened 2016-05-26
Update jdk for travis builds

justinvdk:
Some JavaFX classes are only available in newer versions, so travis needs to update prior building.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,16 @@
 language: java
 jdk:
-        - oraclejdk8
+ - oraclejdk8
+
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,9 +1,16 @@
 language: java
 jdk:
-        - oraclejdk8
+ - oraclejdk8
+
+# Use latest JDK, older version mis some JavaFX classes
```
wouterjsmit: hate to say it, but `mis -> miss`.

justinvdk: fixed

PR #140, opened 2016-05-27
Feat/show meta data

FarisElghlan:
Now the metadata is shown in the user interface when you click on a leaf node.


wouterjsmit:
Looks good, This will be expanded in the next sprint as well.

Issue #62, opened 2016-05-10
PMD generates more errors in the site report than during the build


wouterjsmit:
See https://github.com/ProgrammingLife2016/PL2-2016/issues/58

This is a problem for Jim, because he will be unable to correctly assess us.

PR #144, opened 2016-05-27
Parse and show date of collection

justinvdk:
This PR just checks the content of the cell before assuming it is a date.

### DIFF
--- a/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/Annotation.java
+++ b/PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/Annotation.java
@@ -1,5 +1,6 @@
 package nl.tudelft.pl2016gr2.model;
 
+import java.text.SimpleDateFormat;
 import java.util.Date;
 
 /**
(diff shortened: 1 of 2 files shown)


wouterjsmit:
Not needed for the demo right? Assigning to next sprint then

justinvdk:
You're right. I thought I did? Or maybe no milestone at all. Next week is fine

FarisElghlan:
I guess this can be merged? Code looks good at least, so imo no reason to wait.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/Annotation.java
```diff
@@ -120,7 +121,12 @@ public String buildMetaDataString() {
     sb.append("Age\t\t\t\t\t\t").append(age).append('\n');
     sb.append("Capreomycin\t\t\t\t").append(capreomycin).append('\n');
     sb.append("Cohort\t\t\t\t\t").append(cohort).append('\n');
-    sb.append("Date of collection\t\t\t").append(dateOfCollection).append('\n');
+    sb.append("Date of collection\t\t\t");
+    if (dateOfCollection != null) {
```
wouterjsmit: Both branches are one line long. It is better practice/easier to read to read a positive if-condition than a negative. This is a [style rule imposed by PMD](http://pmd.sourceforge.net/pmd-4.3.0/rules/design.html#ConfusingTernary) (also in our project).

wouterjsmit:
One comment. We all know it is pointless, but it is PMD and we need to also start fixing our site-generated PMD warnings (see #62)

justinvdk:
@Pathemeous fixed it. I think for null-checks the positive version is more confusing because you usually ask yourself "when stuff's alright, go-on".

This is why Java needs to result non-null references to true... (Who cares about the Boolean class anyway ¯_(ツ)_/¯)

PR #146, opened 2016-05-27
Changed the in/out link node IDs

FarisElghlan:
Changed the node IDs which were stored in the in/out edges to actual nodes


wouterjsmit:
Looks good, merging.

PR #149, opened 2016-05-28
Start in correct directory when browsing

### LINKED ISSUE #148: FileChooser starting dir for browse is wrong
When selecting browse in the dropdown I expect to start in the directory of the file that was chosen before. What actually happens is always the working directory of the process.

justinvdk:
fixes #148


wouterjsmit:
It's the one-line diff that made a world of difference

Issue #127, opened 2016-05-23
Genome Strings in GraphNodes consume huge amounts of memory


wouterjsmit:
See https://github.com/ProgrammingLife2016/PL2-2016/pull/124#discussion_r64298321

wouterjsmit:
Moving to milestone 6 because it has no priority for the demo.

PR #161, opened 2016-06-01
Refactor genome storage

wouterjsmit:
The genomes are now stored using integers. This saves a lot of bandwidth. See #127.


FarisElghlan:
Already checked the code/new performance last week, looks good. (from ~500MB to ~120MB if I remember correctly)

FarisElghlan:
The genomemap needs a `clear()` method to clear its content. This must be called at the start of the `parseHeader()` method. Now it is possible to load a new graph file and still keep the old genome mappings (which don't exist anymore in the graph). This causes problems when trying to parse over all genomes in the map (ex. when searching a genome by name).

wouterjsmit:
@FarisElghlan Added.

justinvdk:
As far as I'm concerned we can merge this PR.

PR #167, opened 2016-06-03
Dev new

wouterjsmit:


wouterjsmit:
Combining the previous merges since these were problematic and required conflict resolution in a separate branch

PR #174, opened 2016-06-04
Updated the testing document

FarisElghlan:
Code coverage decreased during the last sprint. This has to be improved this sprint.


wouterjsmit:
Thanks for the update.

PR #184, opened 2016-06-05
Change API for lists in GraphNodes

wouterjsmit:
The lists can now be trimmed.

I deprecated the setter methods for edges. There is no need for them now with the `addAll` methods (which do the same without replacing the list reference inside the node).

I also found a bug in the `removeGenome` method: it removed genomes per index (introduced when switching from genome strings to genome integers).

I also added the trimming actions to all `SequenceNode` constructors, because it was missing in some cases.


FarisElghlan:
Wouldn't it be better to simply remove the set method? Deprecating methods seems like overkill to me, since there isn't some kind of large community using our code, whose code will be broken after removing this method.
Other changes look good.

wouterjsmit:
@FarisElghlan I figured it would be more lenient, especially in the case of `feat/zoom-bubbles`, but on second thoughts you are right. One should just use the new method instead.

wouterjsmit:
Updated. Merging this after checks pass.

Issue #186, opened 2016-06-06
getGenomesOverEdge uses ID but needs constant value


wouterjsmit:
The solution is to use `getLevel`. This can only be fixed when `feat/zoom-bubbles` has been merged.

wouterjsmit:
Applied this in `feat/zoom-bubbles`, after it was merged with the chagnes #187. See #189 .

Issue #169, opened 2016-06-03
SplitGraphs does not set edges in sub graph correctly


wouterjsmit:
The split graphs algorithm only prunes edges that lead to nodes that are no longer in the graph.

This leads to incorrect behaviour.
Example:
Nodes in main graph: `{0, 1, 2}`
Edges in main graph: `{(0 -> 1), (1 -> 2), (0 -> 2)}`
ASCII art:

```
    (1)
   /   \
(0)----(2)
```

Edge `(0 -> 2)` represents genome 1, the path going through node `(1)` represents genome 2.

When a subgraph is created for genome 2, all three nodes still exist in the graph. Therefore, the edge `(0 -> 2)` is not removed.

It should however, because in the subgraph, the path for genome 1 should not exist.

wouterjsmit:
When fixing this I had to introduce a new subtle bug, which must be fixed whenever the semantic zoom branch is ready.

See  #186 for the status of this bug

PR #187, opened 2016-06-06
Fix SplitGraphs bug

wouterjsmit:
Fixes the bug of drawing too many lines for the subset of genomes.

See #169.

To fix this, I corrected the `getGenomesOverEdge` method. This method was rather an estimate than the exact genomes, which I did need however for the SplitGraphs algorithm to select the right edges.

Right now, the method uses IDs, but a more solid value should be used (such as `getLevel()`). This can be done as soon as the semantic branch has been merged.

A ticket for this can be found here: #186

### DIFF
--- a/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/SplitGraphs.java
+++ b/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/SplitGraphs.java
@@ -9,7 +9,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.NoSuchElementException;
 
 /**
@@ -36,9 +35,8 @@ public class SplitGraphs {
   /**
    * Instantiates an algorithmic class on a <code>SequenceGraph</code>.
    * <p>
-   * When {@link #getSubgraph(Collection)} is called, the class will construct a new instance of
-   * the
-   * graph containing only the specified genomes.
+   * When {@link #getSubgraph(Collection)} is called, this class will construct a new instance of
+   * the graph containing only the specified genomes.
    * </p>
    *
    * @param mainGraph The graph to split
@@ -62,9 +60,7 @@ public class SplitGraphs {
     assert mainGraph.getGenomes().containsAll(
         genomeSet) : "Tried splitting graph on absent genomes";
 
-    HashSet<GraphNode> nodeSet = findSubgraphNodes(genomeSet);
-    HashMap<Integer, GraphNode> nodeMap = new HashMap<>();
-    nodeSet.forEach(node -> nodeMap.put(node.getId(), node));
+    HashMap<Integer, GraphNode> nodeMap = findSubgraphNodes(genomeSet);
     return createNewGraph(nodeMap, genomeSet);
   }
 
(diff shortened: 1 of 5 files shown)


PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/SplitGraphs.java
```diff
@@ -89,53 +91,55 @@ public SequenceGraph getSubgraph(Collection<Integer> genomes) {
...
+  private HashMap<Integer, GraphNode> pruneNodes(HashMap<Integer, GraphNode> newNodes,
+      HashMap<Integer, GraphNode> oldNodes, HashSet<Integer> genomes) {
+    newNodes.forEach((id, node) -> {
+      node.addAllInEdges(pruneInEdges(node, newNodes, oldNodes, genomes));
+      node.addAllOutEdges(pruneOutEdges(node, newNodes, oldNodes, genomes));
+      node.addAllGenomes(pruneGenomes(oldNodes.get(id), genomes));
+      if (node.getId() == 8720) {
+        System.out.println("Arrived at node 8720");
+        node.getOutEdges().forEach(System.out::println);
+      }
```
justinvdk: These lines seem be to for debugging purposes.

justinvdk:
I'm not qualified to talk about correctness, but the code seems to be in order.
I didn't see any odd edges in the graph either.

wouterjsmit:
Merging if no further comments?

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/SplitGraphs.java
```diff
@@ -74,68 +73,63 @@ public SequenceGraph getSubgraph(Collection<Integer> genomes) {
...
-
-    newNode.addAllGenomes(pruneGenomes(original, genomeSet));
-    newNode.addAllInEdges(pruneInLinks(original, nodeSet));
-    newNode.addAllOutEdges(pruneOutLinks(original, nodeSet));
+  private HashMap<Integer, GraphNode> pruneNodes(HashMap<Integer, GraphNode> newNodes,
+      HashMap<Integer, GraphNode> oldNodes, HashSet<Integer> genomes) {
+    newNodes.forEach((id, node) -> {
+      node.addAllInEdges(pruneInEdges(node, newNodes, oldNodes, genomes));
+      node.addAllOutEdges(pruneOutEdges(node, newNodes, oldNodes, genomes));
+      node.addAllGenomes(pruneGenomes(oldNodes.get(id), genomes));
```
FarisElghlan, line 129: Shouldn't `set...` be used here instead of `addAll` for all 3 of these?
I think this causes a bug for the genomes, since the copy method coppies the genomes and now you add the genomes to the coppied arraylist, resulting in an arraylist with many duplicates and some non-duplicates which shouldn't have been in the genome list at all.
Note however that the main graph is not altered, since the genome lists are coppied into new arraylist when the node is coppied. Maybe we could change this for some of the lists (if we are sure they aren't altered) so their references are passed (or at least for the `copyAll` methods), since copying all of the lists if they won't be altered is bad for perfomance and memory usage.

FarisElghlan:
Found one possible bug, the rest looks good.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/SplitGraphs.java
wouterjsmit, line 129: I originally meant the `copy()` method to only copy the ID. If that is no longer the case however, you are right and that must be changed.

I removed the `set...()` because I think it is bad practice to allow the caller to specify the collection and its pointer (or the `set...()` methods should remove the old fields and add all of the specified Collection).

I agree that immutable fields would be really nice. I decided not pursue this because of compatibility and limited time. When everything is put together again, we could refactor this.

FarisElghlan, line 129: ok apparently the implementation of the copy method on dev is different from the copy method on the branch which I am currently working on (zoom bubbles)

wouterjsmit, line 129: The zoom bubbles could still adhere to the rule of not copying the genomes,inedges and outedges, as is specified by `GraphNode.copy()`.

I will add tests that assert this is the case to the SequenceNode test suite (which is now at 44 tests).
This suite should be refactored so that the GraphNode interface can be tested for all bubbles. Especially for tests for copying that is useful.

wouterjsmit:
Added tests that verify absence of this bug. Note that the suite should be extended to work for all `GraphNode`s at some point soon.

PR #194, opened 2016-06-08
Drag all genomes in selection from search pane

wouterjsmit:
Before, only one of the genomes (the current row) was dragged.


wouterjsmit:
Merging.

Issue #202, opened 2016-06-08
Graphnode selection is sticky


justinvdk:
When selecting something in the graph, and afterwards select something else (i.e. in the tree).
When the graph is redrawn, the previous selected GRAPH node is selected again. The root of this "bug" is that the selectionManager actually doesn't keep track of the last selected element, but when redrawing the graph you'd want to keep the selected node selected. I hotfixed this by keeping track of the selected node in the SelectionManager but there is no logical point in execution I can find to clear this property, and thus when scrolling the last selected node is always re-selected.

justinvdk:
not applicable anymore

PR #204, opened 2016-06-08
Gui/rootlayout changes

justinvdk:
See #202 too
Some layout reordering required the first small step for the refactor so the SelectionPaneController already exists on this branch, but now doesn't do much as most responsibilities still lie with the SelectionManager.

When code is accepted will clean up commits (or just squash them..)


justinvdk:
Closing for now, still doing some changes and I think its best to do within this PR.

Issue #203, opened 2016-06-08
SelectionManager changes


justinvdk:
We now have a SelectionManager class. I think this class is getting too many responsibilities and we should introduce a new class, e.g. SelectionPaneController with a corresponding FXML file. We can than drop the fxml into the layout somewhere and set it up in the rootLayoutController.java.

Then a refactor is needed to eventually remove SelectionManager.
Things to consider:
- #202
- Where do we now keep track of the top/bottom Genomes)
- Can we only select a single thing at a time, or can we select 1 thing in the tree + 1 thing in the graph etc.
- Consistent naming of FXML's
- Probably more

justinvdk:
New plan:
SelectionManager stays, but in a much more lighter form.
Selection isn't that easy as it sounds, there are different types and usually the actor in selecting (a visual node) does not share a similar lifecycle with its data object.  

GraphPaneController and TreePaneController are now responsible for their own selection.
This might seem odd, please do share your opinion but this is the challenge:

A lot of "actors" in selection, i.e. the JavaFX Node's, do not have a similar lifecycles with their corresponding data nodes. And thus there is no good way to reference back to have the SelectionManager KNOW whether a given (JavaFX) Node is selected. There ARE ways, but since they differ for graphnodes and treenodes it gets complicated.

For now this is my (hacky) solution:
ISelectionInfo now includes a 'getSelectionIdentifier' method which returns an Object. The idea is that the object will be used to check for equality. The graphNodes return their corresponding datanodes, and the treeNodes theirs. This way the JavaFX panes can be redrawn many times without losing the information about what is selected, and a way to highlight these JavaFX Nodes when one appears to correspond to a selected item.

justinvdk:
Furthermore, the new SelectionManager knows 'categories', for now only Graph and Tree (and I think it will probably stay this way). For each category the user is able to multi-select items.

justinvdk:
See related commit. @Pathemeous please don't hate :(

justinvdk:
```
              +--->selectedGraphNodes
              +--->selectedTreeNodes
              |
          +----------------+
     +--->+SelectionManager+<---+
     |    +----------------+    |
     |           ^              |
     v           |              v
  TreePane       |         SelectionPane
  Controller     v         Controller
              GraphPane
              Controller
```

So every component makes sure it notifies the SelectionManager when something is selected/deselected. The selectionManager keeps track. Every component is then allowed to listen for changes and update their view accordingly.

[...]

PR #206, opened 2016-06-09
Embed search and selectionPane into rootLayout

justinvdk:
Make ctrl+f request focus and esc un-focus search
Start refactor of SelectionManager
Make returned node more flexible
TreeLeaf- and TreeNodeDescription now return a more flexible javaFX Node
Change graphNode description pane FXML
Removed unreliable test
Make use of fx:include for the root layout
Make FXMLLoader usage somewhat easier to read
Rename treeManager -> TreePaneController
This is more consistent with the member names and fxml ids
Rename treeManager -> TreePaneController
This is more consistent with the member names and fxml ids
Make graphPaneController responsible for its own nodes

In short: rootlayout now embeds search pane and selection info pane.


justinvdk:
See #203. Will open PR for that when this one is merged.

cbilstra:
From what I can see a lot of small changes, looks good.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/selection/SelectionPaneController.java
```diff
@@ -0,0 +1,49 @@
...
+    rootPane.setBackground(new Background(
+        new BackgroundFill(BACKGROUND_COLOR, null, Insets.EMPTY)));
+  }
+
+  /**
+   * Sets the content of the selectionPane.
+   *
+   * <p>
+   *   This removes previous shown content,
+   *   call with null to clear.
```
wouterjsmit: Please remove the indents here. IntelliJ makes them automatically but shouldn't be there.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/selection/SelectionPaneController.java
```diff
@@ -0,0 +1,49 @@
...
+   * </p>
+   *
+   * @param content new content, may be null
+   */
+  public void setContent(Node content) {
+    rootPane.getChildren().clear();
+    if (content != null) {
```
wouterjsmit, line 40: This is simply fugly. The whole method's functionality is inside the not-null condition.

You could use @nullable or something to denote that it may be `null` (not sure if that's an IntelliJ or JavaDoc feature though).

Why may it be null? IMO the caller should separately use something like `clearContent()` to remove the content. Not entirely sure where this is called, though. This separates the clearing part (outside of the condition) from the fillling part (inside of the condition).

justinvdk, line 40: I'll introduce a clearContent method for now. It emerged like this because setContent was already clearing the view anyway.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/selection/SelectionPaneController.java
justinvdk: Odd, I've changed this in a commit but the preview still shows the old code?

wouterjsmit:
Changes in this PR look good. The test that you removed is unstable, but Faris also removed it. One of you will have to rebase against the other and remove his commit that tries to redo the action. In this case, probably Faris

wouterjsmit:
Ready for Merging I think?

PR #208, opened 2016-06-09
Feat/create annotation reader

cbilstra:
I think AbstractNode is not used anymore, so I commented it. The same for SequenceNodeTest. Furthermore, this PR contains an AnnotationReader and Annotation class

### DIFF
--- a/PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/AnnotationReader.java
+++ b/PL2/PL2-parser/src/main/java/nl/tudelft/pl2016gr2/parser/controller/AnnotationReader.java
@@ -1,6 +1,7 @@
 package nl.tudelft.pl2016gr2.parser.controller;
 
 import nl.tudelft.pl2016gr2.model.Annotation;
+import nl.tudelft.pl2016gr2.model.GenomeMap;
 import nl.tudelft.pl2016gr2.util.Pair;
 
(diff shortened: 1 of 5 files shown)


FarisElghlan:
AbstractNode was removed, so you can remove those files, but why did you comment sequencenodetest?

wouterjsmit:
You did something really weird with Git here.

First of all: remove commit 62e88d2 (`Fix checkstyle?`). It's not necessary, that code passed Checkstyle earlier and you did not change it.

The only checkstyle commit for your own work is 3c67464 (`this must fix checkstyle or I will cry`).

I'm not sure what happened in `98f732a` (`Fix merge conflicts`). I am guessing that you had an outdated local branch and pulled the remote into it **after** making your own commits.

To fix all this, I think that you need to do:

```
git fetch origin
git checkout feat/create-annotation-reader
git rebase -i origin/dev
```

In the rebasing window, do the following.
- Remove the line with the `Fix checkstyle?` commit.
- Remove the line with the 'fix merge conflicts' commit.

You might have to resolve conflicts, although that shouldn't be the case if you only touched your own new files. (thus it being really weird that you have a commit about fixing merge conflicts that hits 106 files).

Edit: Also uncomment SequenceNodeTest please, as Faris asked. You can do this after the rebase.

wouterjsmit:
Great, merging when checks are done.

PR #209, opened 2016-06-09
Bug/zoom bubbles

cathmer:
This makes the phylo bubbles compatible again with the other bubbles.

I've done this by removing the first and last node of every straight sequence. 

EDIT: I did try to squash the commits, but it gave me an error indicating that the rebase could not be applied.. I think it is because I merged dev into my branch in between.


PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/graph/nodes/StraightSequenceBubble.java
```diff
@@ -37,12 +38,23 @@ public GraphNode copy() {
...
+  
+  @Override
+  public Collection<Integer> getGenomes() {
+    HashSet<Integer> genomeSet = new HashSet<>();
+    for (GraphNode nestedNode : getChildren()) {
+      genomeSet.addAll(nestedNode.getGenomes());
+    }
+    for (GraphNode inEdge : getInEdges()) {
+      genomeSet.addAll(inEdge.getGenomes());
+    }
```
FarisElghlan, line 47: Why are the genomes of the inedges needed here? This doens't really seem correct.
If this is really needed you could also let the parent class (Bubble) return a HashSet instead of a Collection (this is a child class of Collection, so it still overrides the method correctly) and call
`HashSet<Integer> genomeSet = super.getGenomes()`
that way you can remove these lines:

``` java
for (GraphNode nestedNode : getChildren()) {
   genomeSet.addAll(nestedNode.getGenomes());
}
```

so there will be less code duplication. This is not really important but just looks a bit nicer.

FarisElghlan:
Found one method which looks a bit strange, the rest looks good. I do think that the changes to `StraightSequence` are not really needed however, but it's good enough for now.

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/graph/nodes/StraightSequenceBubble.java
cathmer, line 47: It is needed because the way the straight sequence is implemented now (without the first and last node) some genomes can be lost in the sequence bubble itself when its nested node is an indel. Because in this case it's still perceived as a straight sequence, but in reality it's not because there is an edge directly from the start to end node. So I do need to include the genomes of the node before for the filtering algorithm to work correctly. 

I think I could actually just add the genomes of the startnode, since all genomes that pass through there will also be in the rest of the sequence.

cathmer:
I removed the unnecessary code of the getGenomes() method.

PR #210, opened 2016-06-09
Fixed bug when there are multiple root nodes

FarisElghlan:
Also changed the drop event to not set the completed status to true,
so when dragging from a text editor the text will remain there.

Now the graph of 328 genomes can be drawn without bugs. Next todo: speed up the drawing of edges.

### DIFF
--- a/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/SubgraphAlgorithmManager.java
+++ b/PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/SubgraphAlgorithmManager.java
@@ -21,6 +21,8 @@ import java.util.logging.Logger;
  */
 public class SubgraphAlgorithmManager {
 
+  private static int dummyRootNodeId = Integer.MAX_VALUE;
+
   /**
    * This is a utility class, so let no one create an instance of it.
    */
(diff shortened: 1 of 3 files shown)


PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/SubgraphAlgorithmManager.java
```diff
@@ -21,6 +21,8 @@
  */
 public class SubgraphAlgorithmManager {
 
+  private static int dummyRootNodeId = Integer.MAX_VALUE;
```
wouterjsmit, line 24: Aren't we running out of Integer ranges to assign unique IDs?

wouterjsmit:
Do we not know why the root nodes bug? It seems very fragile to just put a dummy node in front of it.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/SubgraphAlgorithmManager.java
FarisElghlan, line 24: We can still use `Integer.MAX_VALUE / 2 ++`, `Integer.MAX_VALUE / 2 -1 --` and the same for `MIN_VALUE` :)

wouterjsmit:
Right, quick fix is quick

PR #211, opened 2016-06-09
Show ellipses for SequenceNode when too long

justinvdk:
The ellipses were missing and thus it wouldn't be visible that the sequence is actually longer.


FarisElghlan:
Looks good!

wouterjsmit:
Excellence is in the details.

PR #212, opened 2016-06-09
 Fixed the edge drawing performance

### LINKED ISSUE #182: Edge drawing optimization
Investigate if it is possible to draw the edges more efficiently, or otherwise draw fewer edges.

The application zooms/scrolls very fast when the edges are not drawn. When the edges are drawn this is not always the case (especially for larger datasets), because there can be 100's of thousands of edges at the same time and currently a separate JavaFX Line object is created for each edge. This can possibly be done more efficient if something like a Canvas is used, where it is not necessary to create a new object for every drawn line.

FarisElghlan:
Fixes #182

The edges are now drawn in a Canvas, resulting in lower memory
usage and faster drawing speeds.
The edge width is now approximated, so this no longer is a bottleneck
The 328 graph can now be redrawn at around 60 FPS on my PC.
The lowest FPS rate I got was around 20-25.

### DIFF
--- a/PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/GraphPaneController.java
+++ b/PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/GraphPaneController.java
@@ -10,6 +10,8 @@ import javafx.collections.ObservableSet;
 import javafx.event.ActionEvent;
 import javafx.fxml.FXML;
 import javafx.fxml.Initializable;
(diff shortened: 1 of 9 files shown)


wouterjsmit:
Please rebase against `dev` since that commit is now on `dev`.

FarisElghlan:
Fixed (I think/hope)

wouterjsmit:
Did you try it on the 340/328 graph?

Stuff seems pretty overloaded

wouterjsmit:
Not sure what goes wrong with the 328 data set for me. but the optimizaiton can be merged.

PR #214, opened 2016-06-09
Added shared module tests

FarisElghlan:
Added some tests


justinvdk:
Tests are always welcome. I think they can be merged as soon as they pass on Travis.

PR #223, opened 2016-06-12
Feat/heatmap options

### LINKED ISSUE #142: Application needs heatmap/visualizaton of the drug resitance

FarisElghlan:
Added a second heatmap and the possibility to change the value value which is displayed in the heatmap.

Fixes #142


cbilstra:
This PR contains a lot changes which are also in the other PR (#224) or am I wrong?

FarisElghlan:
#224 was made based on this branch

justinvdk:
Code looks good. Only thing that caught my eye is the Iterator type of `IPhylogeneticTreeNode`.
Is this to prevent confusion with `PhylogeneticTreeNode` implementing the `Iteratable` interface himself (and thus having two `iterator()` methods?)

FarisElghlan:
You can't implement the same interface twice with different generic types. If the `IPhylogeneticTreeNode` interface would extend `Iterable<IPhylogeneticTreeNode>` then the `PhylogeneticTreeNode` implementation of the interface would have to return an iterator with the interface as generic type, thus not allowing the `PhylogeneticTreeNode` implementation to get an iterator of its child nodes with the correct generic type.

justinvdk:
I see. Wasn't sure whether java did support this.

PR #225, opened 2016-06-14
Implement Random Bubbling algorithm

cathmer:
With this algorithm RandomBubbles are made which are solely based on the graph, and not on phylogeny, as was one of the wishes of the client. 

Right now the algorithm is not used anywhere, because changes in the GUI need to be made to reflect the different type of Bubble. This is a task for @FarisElghlan.


FarisElghlan:
Maybe GraphBubble would be a beter name?

cathmer:
Renamed RandomBubble to GraphBubble.

cbilstra:
Looks good.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
```diff
@@ -1,38 +1,27 @@
...
-import java.util.Set;
 
 /**
  * Class used to filter bubbles based on the phylogenetic tree.
  *
  * @author Casper
  *
  */
-public class FilterBubbles implements PhyloFilter {
+public abstract class FilterBubbles implements PhyloFilter {
```
wouterjsmit: How about `BubbleFilter` (rather than a verb 'to filter bubbles' it is a filter (noun) for bubbles)

Also update the class JavaDoc

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
```diff
@@ -57,189 +46,34 @@ public FilterBubbles(Collection<GraphNode> orderedNodes) {
       originalOutEdges.put(node.getId(), new ArrayList<>(node.getOutEdges()));
     }
   }
-
+
```
wouterjsmit, line 29: You inserted trailing whitespace here.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
```diff
@@ -57,189 +46,34 @@ public FilterBubbles(Collection<GraphNode> orderedNodes) {
...
-    orderedNodes.removeAll(bubble.getInEdges());
-
-    CompareSubgraphs.alignVertically(orderedNodes, bubble.getInEdges());
-
-    return orderedNodes;
+  protected Map<Integer, Collection<GraphNode>> getOriginalInEdges() {
+    return originalInEdges;
   }
-
+
```
wouterjsmit, line 38: please fix all whitespace in your files

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
wouterjsmit: Rename `PhyloFilter` to `BubbleFilter` or `Filter` and call this class `Abstract<...>`

Also update the documentation for the old `PhyloFilter` interface.

Add the relevant methods (you specify some generic methods in this abstract class) to the interface rather than the abstract implementation (unless these are inherently different from other `PhyloFilter`s, which I don't think is the case)

wouterjsmit:
A comment that I put on a single line first but is worth mentioning for the whole PR.

> Rename `PhyloFilter` to `BubbleFilter` or `Filter` and call this class `Abstract<...>`
> 
> Also update the documentation for the old `PhyloFilter` interface.
> 
> Add the relevant methods (you specify some generic methods in this abstract class) to the interface rather than the abstract implementation (unless these are inherently different from other `PhyloFilter`s, which I don't think is the case)

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
```diff
@@ -57,189 +46,34 @@ public FilterBubbles(Collection<GraphNode> orderedNodes) {
...
-      for (Integer leaf : leaves) {
-        if (genomes.contains(leaf)) {
-          curNodeOutlinks.add(outlink);
-          break;
-        }
-      }
-    }
-
-    return curNodeOutlinks;
+  protected Collection<GraphNode> getRootNodes() {
```
wouterjsmit: You are storing the root nodes of the `SequenceGraph` in this object?

That is duplication of storage. How do you know that the stored nodes here will remain correct for the whole graph?

A node can be checked to be a rootNode easily with `node.getInEdges().isEmpty()`, this is the same computation cost as `FilterBubble.getRootNodes().contains(node)`, possibly even faster, depending on the `Collection()` implementation.

Edit: If you need the rootNodes from the graph, save the whole SequenceGraph instead. It is more logical to instantiate a filter on a graph (SplitGraphs does the same).

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterGraphBubbles.java
```diff
@@ -0,0 +1,160 @@
...
+    if (bubble != null) {
+      for (GraphNode node : nestedNodes) {
+        bubble.addChild(node);
+      }
+    }
+    
+    return bubble;
+  }
+  
+  private Bubble makeBubble(Queue<GraphNode> toVisit, Set<GraphNode> visited, GraphNode start,
```
wouterjsmit: Add comments please (Jim wants this)

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterGraphBubbles.java
```diff
@@ -0,0 +1,160 @@
...
+          poppedNodes.add(newBubble);
+          newBubbles.add(newBubble);
+        }
+        FilterHelpers.addToVisit(newBubble.getOutEdges().iterator().next(), toVisit, visited);
+      }
+    }
+    return poppedNodes;
+  }
+  
+  private Bubble createBubble(GraphNode start, GraphNode currentEnd) {
```
wouterjsmit: Add comments please (Jim wants this)

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/ZoomInGraph.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Class to zoom in on random bubbles.
+ * 
+ * @author Casper
+ *
+ */
+public class ZoomInGraph extends ZoomIn {
```
wouterjsmit: Really weird name. This is the ZoomIn implementation for 'normal' graph bubbles?

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/subgraph/CompareSubgraphs.java
```diff
@@ -167,7 +168,7 @@ private static ComplexVerticalArea calculateGraphArea(GraphNode node,
     areaMap.put(node, complexNodeArea);
 
     SimpleVerticalArea nodeArea = complexNodeArea.getLargestArea();
-
+
```
wouterjsmit: Inserted trailing whitespace

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/graph/nodes/AbstractGraphNode.java
```diff
@@ -154,7 +154,7 @@ public void trimToSize() {
     //inEdges.trimToSize();
     //outEdges.trimToSize();
   }
-
+
```
wouterjsmit: Trailing whitespace

PL2/PL2-shared/src/main/java/nl/tudelft/pl2016gr2/model/graph/nodes/GraphBubble.java
```diff
@@ -0,0 +1,79 @@
...
+
+import nl.tudelft.pl2016gr2.visitor.NodeVisitor;
+
+import java.util.Collection;
+import java.util.List;
+
+public class GraphBubble extends AbstractGraphBubble {
+
+  /**
+   * Construct a phylo bubble.
```
wouterjsmit: Update all javadoc in this class

wouterjsmit:
I reviewed your PR.

Please make sure to check all JavaDoc while processing my feedback. You did some copying and some JavaDoc is completely out of place now.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/FilterBubbles.java
cathmer: I don't have access to the SequenceGraph, so I can't store it. 

I don't see why storing the root nodes is a problem, because they won't change (because the root nodes of the graph before it's been filtered by this class are stored). And I don't need to check whether nodes are a root node, I just need the root nodes for my algorithm. 

I could instead save the orderedNodes list that is given is as input and then calculate the rootNodes on the fly, but that would mean that I would need to save this list. Also, this list could change, so basing the rootNodes on this would not be very secure.

PL2/PL2-core/src/main/java/nl/tudelft/pl2016gr2/core/algorithms/ZoomInGraph.java
cathmer: Changed it to GraphBubbleZoom.

cathmer:
Renamed some classes, updated the JavaDoc and removed some trailing whitespaces.

wouterjsmit:
Looks good, merging

PR #226, opened 2016-06-14
Fact/styles to css

FarisElghlan:
Refactored the graph/tree/legend to use the CSS instead of hardcoded colors. Also did some color updating.
The search pane will be refactored when the code is done.


wouterjsmit:
Great, looks clean. More deletions and additions, which means it is a good refactor

PR #235, opened 2016-06-15
Moved the bubble algorithms to packages

FarisElghlan:
Moved the bubble algorithms to packages and deleted unused files


wouterjsmit:
Clean

PR #236, opened 2016-06-15
Added an algorithm chooser

### LINKED ISSUE #232: Change bubble algoritm
Make a button to change the bubbling algoritm which is used.

FarisElghlan:
It is now possible to select which bubbling algorithms
have to be performed on the graph, see settings menu.

Fixes #232


cbilstra:
Wouldn't it be nicer if the user could specify how often some bubbling algorithm is chosen and in which order? Looking at the first file changed the order is static and each algorithm can only be run once

FarisElghlan:
Yes, but I there are still around 22 hours worth of issues left which I have to do within two days

cathmer:
I don't think it's useful to run the algorithm multiple times, because it's not like point mutations and indels can be detected multiple times. 

However, I do think it would be nicer if it's possible to select point mutations/indels/straight sequences together with graph bubbles as well, and not just with phylogenetic bubbles. Because right now this decreases the value of the graph bubbles heavily.

Also, when I select indels together with phylogenetic bubbles something goes wrong, because the upper branch of the big bubble in the graph of 10 stops really early.

EDIT: Don't forget to add Graph Bubbles to the legend.

FarisElghlan:
Indels can actually be detected multiple times, since you can find an indel within an indel (although I'm not sure if in this case the outer indel is still called an indel).
I will adjust it so indels/point mutations and straight sequence bubbles can also be selected when filtering based on the graph.

FarisElghlan:
Merge this after #238 has been merged into this branch.

wouterjsmit:
Looks good

justinvdk:
I see #238 is merged. Is this branch ready to be merge?

FarisElghlan:
yes

Issue #230, opened 2016-06-14
Improve the description pane


FarisElghlan:
Add more information to the description panes and improve their usability (by making them bigger and moving the copy button).

justinvdk:
Implemented a description pane for all nodes currently implemented.

Currently not done:
- [ ] button to highlight/select TreeNode from description pane of Phylogenetic bubble.
- [ ] improve button placement of copy.

PR #237, opened 2016-06-15
Feat/selection description implementations

justinvdk:
Relates to #230
Graph description panes are a bit more sensible now.
Highlighting / relation to tree is not implemented in this PR


PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/selection/GraphBubbleDescriptionBuilder.java
```diff
@@ -0,0 +1,112 @@
...
+ * <p>
+ * The internal logic uses the visitor-pattern to provide a fitting result for each
+ * given {@link GraphNode}.
+ * </p>
+ * <p>
+ * Please see {@link #buildInfo(GraphNode)} when you want to use this class.
+ * </p>
+ *
+ */
+public class GraphBubbleDescriptionBuilder implements NodeVisitor {
```
wouterjsmit, line 27: Not sure this is the original idea behind `NodeVisitor` objects, but I guess the pattern is very flexible?

Edit: disregard, I didn't understand correctly what this class was.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/selection/PointMutationBubbleDescriptionController.java
```diff
@@ -0,0 +1,94 @@
...
+   *
+   * @param bubble the {@link PointMutationBubble} that holds the data.
+   */
+  public void setup(PointMutationBubble bubble) {
+    if (bubble.getChildren().size() == 2) {
+      Iterator<GraphNode>  it = bubble.getChildren().iterator();
+      GraphNode childOne = it.next();
+      GraphNode childTwo = it.next();
+
+      if (childOne.getClass() == SequenceNode.class && childTwo.getClass() == SequenceNode.class) {
```
wouterjsmit: So what if they aren't `SequenceNode`s?

You can use `node.hasChildren()` to determine whether it is a root-level node (aka is not a bubble)

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/selection/SelectionPaneController.java
```diff
@@ -55,16 +59,37 @@
...
+   * <p>
+   * This method calls {@link ISelectionInfo#getNode()} on the given {@link ISelectionInfo}
+   * and catches its {@link IOException}. In the latter case it will return a simple
+   * {@link Text} with an error.
+   * </p>
+   *
+   * @param info the given {@link ISelectionInfo}
+   * @return a {@link Node} that can be displayed
+   */
+  private Node safeGetDescription(ISelectionInfo info) {
```
wouterjsmit, line 74: defensive programming is good but I still cry that this is necessary (if it is)

wouterjsmit:
General structure is very organized. Props.

Small comments:
JavaDoc paragraphs are usually newlined after each sentence. This is a maintainabiltiy thing (so that changes to one sentence will not span the line of a partial other sentence).

Is there no way to skip the exception for the `getNode`?

Also, the `getClass()` is not really desireable, I proposed alternatives.

FarisElghlan:
Did you already add the lineage colors in the description? I think it is also more clear to use the name than just the color (or use the name with the color as background or use it as text color or something like that)

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/selection/PointMutationBubbleDescriptionController.java
justinvdk: My understanding is that `SequenceNode`s are the only possible children for this particular node. If not it wouldn't be a pointmutation. Correct me if I'm wrong though!

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/selection/SelectionPaneController.java
justinvdk, line 74: The thing is, the FXMLLoader will throw an IOException which in java the programmer SHOULD always catch as these exceptions simply might occur at runtime. 
As the only reasonable case in which this exception is thrown, the cause is most likely that the `.fxml` file does not exist. Other scenario's would be actual exceptions / errors with IO. When that kind of stuff goes wrong it is not worth it for our application to do something sensible besides not directly crashing. (IMHO)

justinvdk:
@Pathemeous 

> JavaDoc paragraphs are usually newlined after each sentence. This is a maintainabiltiy thing (so that changes to one sentence will not span the line of a partial other sentence).

Do you mean a newline, still inside the `<p>` tag? Like so on line 4:

```
1. <p>
2. lorem ipsum bla bla
3. more bla
4. 
5. </p>
```

> Is there no way to skip the exception for the `getNode`?

I thought about this. At first I added it to prevent a lot of try-catch'es in locations that would be really cumbersome. Now that you ask I actually thing it is the proper location.The method `getNode` itself shouldn't be responsible for catching that particular exception. We're asking it to give us a (javafx) `Node` to display to the user. The `safeGetDescription` is there to have a proper structure even though there is not enough time to implement it properly (because as of now all we do is give a `Text` with an error message, the user should actually be informed of an error in a proper way).

> Also, the getClass() is not really desireable, I proposed alternatives.

Do you have any concrete alternatives? I tried to work around this but my decision was based on effort vs reward. In the cases that the children would not be that particular class there was nothing sensible for the `Pane` to show (I think, if you think otherwise please tell).

@FarisElghlan 

> Did you already add the lineage colors in the description? I think it is also more clear to use the name than just the color (or use the name with the color as background or use it as text color or something like that)

No, good point. ~~As of now the Collection of `MetaData` (why is the D capitalized?) is not readily available near the code of the descriptions. This would mean expanding the `ISelectionInfo` even more or adding it to the underlying data objects, both of which I think can get really messy. I will look into this.~~
I was reviewing #236, which added the metadata to the `GenomeMap` singleton. With that it will be possbile to grab the lineages.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/selection/PointMutationBubbleDescriptionController.java
wouterjsmit: It should be true that they are non-bubbles, but that would be better to check with `hasChildren`. If we have another root-level node implementation, this breaks (We nearly had a ConvergentEvolutionNode, so it can happen)

wouterjsmit:
@justinvdk 

```
/**
 * <p>
 * First sentence.
 * Second sentence.
 * </p>
 */
```

When a sentence simply cannot span only 1 line, then you can wrap it (but don't start a new sentence on an existing line)

`GraphNode` offers `hasChildren()` to determine if it is a composite / leaf object. You are looking for leaves only, meaning that you can use  `hasChildren() == false` to determine whether it is a `Node` type, which offers the `getSequence()` method.

justinvdk:
I took a look at the comments.

Also implemented lineage colors.

wouterjsmit:
You da best.

PR #242, opened 2016-06-16
Feat/graph heatmap

### LINKED ISSUE #228: Graph heatmap
Create a heatmap for the graph. The possible options which can be displayed in the heatmap are:
- the amount of mutations of a certain type (ex. indel or indel and straight sequence)
- the equalities between the top and bottom graph (optional)
- ???

FarisElghlan:
Added a heatmap to the graph.

Fixes #228

### DIFF
--- a/PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/GraphPaneController.java
+++ b/PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/GraphPaneController.java
@@ -652,14 +652,14 @@ public class GraphPaneController implements Initializable {
     updateGraphSize();
     graphUpdater.update();
   }
-  
+
   /**
    * Clear the canvas of the edges.
    */
   private void clearEdgeCanvas() {
-    topEdgeCanvas.getGraphicsContext2D().clearRect(0, 0, topEdgeCanvas.getWidth(), 
+    topEdgeCanvas.getGraphicsContext2D().clearRect(0, 0, topEdgeCanvas.getWidth(),
         topEdgeCanvas.getWidth());
-    bottomEdgeCanvas.getGraphicsContext2D().clearRect(0, 0, bottomEdgeCanvas.getWidth(), 
+    bottomEdgeCanvas.getGraphicsContext2D().clearRect(0, 0, bottomEdgeCanvas.getWidth(),
         bottomEdgeCanvas.getWidth());
   }
 
@@ -908,6 +908,11 @@ public class GraphPaneController implements Initializable {
     double startY = fromNode.getGuiData().relativeYPos * fromRange.rangeHeight
         + fromRange.rangeStartY;
     double endY = toNode.getGuiData().relativeYPos * toRange.rangeHeight + toRange.rangeStartY;
+    if (toNode.getGenomeSize() < fromNode.getGenomeSize()) {
+      graphicContext.setStroke(toNode.getMostFrequentLineage().getColor());
+    } else {
+      graphicContext.setStroke(fromNode.getMostFrequentLineage().getColor());
+    }
     graphicContext.strokeLine(startX, startY, endX, endY);
   }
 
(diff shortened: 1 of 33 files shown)


wouterjsmit:
Please rebase against `dev`

FarisElghlan:
Merged dev into this branch.

wouterjsmit:
All commits except the last are already part of dev. Could you in one way or another remove those ocmmits from this history?

FarisElghlan:
Is that really important? they aren't shown in the diff anymore and I don't really know how to remove them (besides removing this branch, creating a new branch and cherry picking the changes).

wouterjsmit:
Since I'm curious we'll see if Git will actually filter them

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/GraphPaneController.java
```diff
@@ -156,10 +182,50 @@ public void initialize(URL location, ResourceBundle resources) {
...
+      graphUpdater.update();
+    });
+  }
+
+  /**
+   * Add all of the possible heatmap options.
+   */
+  private void addHeatmapOptions() {
+    GraphicsContext heatmapGraphics = heatmap.getGraphicsContext2D();
+    heatmapOptions.put("Amount of mutations (parralel paths in the graph)",
```
wouterjsmit: `parralel -> parallel`

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/heatmap/AbstractHeatmap.java
```diff
@@ -0,0 +1,120 @@
...
+    double startX = endX - nodeWidth;
+    heatmapGraphics.setFill(color);
+    heatmapGraphics.fillRect(startX, 0, endX - startX, HALF_HEATMAP_HEIGHT * 2.0);
+
+    color = DEFAULT_COLOR;
+  }
+
+  @Override
+  public void visit(GraphNode node) {
+  }
```
wouterjsmit, line 74: Do insert something in terms of logging/ assertions to verify that this base version never gets called.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/heatmap/AbstractHeatmap.java
```diff
@@ -0,0 +1,120 @@
...
+    colorChildren(bubble);
+  }
+
+  @Override
+  public void visit(IndelBubble bubble) {
+    colorChildren(bubble);
+  }
+
+  @Override
+  public void visit(GraphBubble bubble) {
```
wouterjsmit, line 97: The visits are identical for every bubble type. Why differentiate?

FarisElghlan, line 97: They are overwritten in the subclasses. This way indel bubble only has to overwrite 1 method for example:

``` java
public class IndelDensity extends AbstractHeatmap {

  public IndelDensity(GraphicsContext heatmapGraphics, ObservableDoubleValue zoomFactor) {
    super(heatmapGraphics, zoomFactor);
  }

  @Override
  public void visit(IndelBubble bubble) {
    setColor(MUTATION_COLOR);
  }

  @Override
  protected double getWidthMultiplier() {
    return 2.0;
  }
}
```

wouterjsmit, line 97: But you could simply not define them at all in the abstract class then right? At runtime the correct heatmap is selected, so the overloaded method is available.

FarisElghlan, line 97: No, all of the subclasses overwrite a different method, for example the `PhyloBubbleDensity` class overwrites the `void visit(PhyloBubble bubble)` method, while the `PointMutationDensity` class overwrites the `void visit(PointMutationBubble bubble)` method. All methods which are not overriden in the subclasses need to behave as specified in the parent class.

wouterjsmit, line 97: Ah, I see your point. Then I agree

PR #244, opened 2016-06-16
Add additional line to show highligh in tree

justinvdk:
JavaFX lines do not support multiple colorings.
Hacks with LinearGradient are not feasable, so added an additional line that
will be used for the inner color


wouterjsmit:
Looks good

PR #248, opened 2016-06-17
Added a minimum scroll delta to scroll more smoothly on macs

FarisElghlan:


wouterjsmit:
Great.

Let's, for simplicity, assume that -0.03% coverage is a measurement error by coveralls.

PR #253, opened 2016-06-17
Annotation bugfix and extra heatmap

FarisElghlan:
Fixed some small bugs before the presentation


justinvdk:
I agree with those changes. Seen the bugs firsthand and think they should be included.
NOTE: also adds an extra heatmap option for the graph, which technically is a new feature. (hence the gui tag)

wouterjsmit:
I don't want to become nazi again, but I kinda do.

Could you separate your commits to reflect the bug fixes. Also ahve a commit message that explains what was wrong (why was the change necessary and why the way you implemented it)

Otherwise I'm sure your code is correct.

wouterjsmit:
Please do this since I want to merge your code

wouterjsmit:
Thanks, merging

FarisElghlan:
...

FarisElghlan:
?

wouterjsmit:
Sorry, waiting for checks but then had to travel.

PR #260, opened 2016-06-18
Fact/compare performance

FarisElghlan:
Bubbling the graph of 328 genomes now takes around 3-6 seconds instead of around 14-18 seconds. Many operations have been changed from either:
O(n<sup>2</sup>) to `O(n)`
`O(n)` to `O(log(n))`
or
`O(n)` to `O(1)`

I expect the performance difference will be a lot larger if the dataset gets larger. Most of the improved operations have the number of genomes as a factor for `n`. Since 328 is not a large number the performance gain wasn't that large for this dataset.
The vertical alignment is now the most expensive operation which is performed when the graph is calculated.


wouterjsmit:
Great improvement! Waiting for #253 and a rebase on dev before merging (otherwise I"m fine with a rebase on `dev` without 0258f3e if that doesn't cause merge conflicts.

justinvdk:
#253 is merged. @FarisElghlan will merge this one after a rebase onto dev+checks (I promise..)

FarisElghlan:
rebased

PR #261, opened 2016-06-19
Feat/improve vertical spacing

FarisElghlan:
The amount of vertical space which each path through the graph gets now depends on the amount of genomes which go through the path.
Each path gets 20% of the space by default and the rest of the vertical space is divided according to the amount of genomes.

Before:
before

After:
after

Still not perfect, but a lot better than before (especially when vertical zooming is implemented I think the new version will be a lot more usefull).


FarisElghlan:
Closing because there still are some bugs...

PR #262, opened 2016-06-19
Feat/improve vertical spacing

FarisElghlan:
The amount of vertical space which each path through the graph gets now depends on the amount of genomes which go through the path.
Each path gets 25% of the space by default and the rest of the vertical space is divided according to the amount of genomes.

Before:
before changes

After:
after changes

The path are sorted from top to bottom with at the top the path with the least genomes and at the bottom the path with the most genomes. This means the most common path isn't in the center anymore, but this is the easiest way to avoid a lot of edge collisions.


FarisElghlan:
Coincidental bonus feature:
Long vertical edges now indicate few genomes go through the path.
Shorter vertical edges indicate many genomes go though the path.

justinvdk:
@FarisElghlan Lots of changes that will be merged via other PR's.

Could you rebase this onto dev after #260 is merged, I will then re-review (which hopefully will be much less changes ;) )

wouterjsmit:
Are you done working on this @FarisElghlan?

FarisElghlan:
yes

PR #276, opened 2016-06-20
Maximize window on startup

### LINKED ISSUE #269: Maximize on start up [Could have]
Maximize on start up by default.

justinvdk:
improved ux a bit
closes #269


wouterjsmit:
Good, can't really be wrong can it.

PR #280, opened 2016-06-20
Set minimum width of button to prevent ellipses

justinvdk:


wouterjsmit:
Great, fixes regression from #278

PR #281, opened 2016-06-21
Removed unnecessary references to hashgraph

FarisElghlan:
A reference to a hashgraph was kept in a variable, while this is unnecessary.
I have removed the reference. This way the garbage collector can free the memory.


wouterjsmit:
Looks good. Merging.

PR #282, opened 2016-06-21
Feat/annotation arrows

### LINKED ISSUE #257: Forward and Reverse annotations are not differentiated
There is a field that denotes the direction of the gene in the Gene Annotation file. This field contains a `+` for 'forward'  genes and a `-` for 'reverse' genes.

A forward gene is a gene that is present on one side of the DNA helix, a reverse is one that is present on the other side.

Both can be relevant, and the direction (apparently) makes a difference.

The customer proposed adding a forward or reverse arrow to subtly note the difference.

FarisElghlan:
Added arrows to the annotations to indicate their direction.

Fixes #257


FarisElghlan:
Example:
arrows

wouterjsmit:
Please rebase on `dev`.

wouterjsmit:
While rebasing, also edit the arrow commit's message to include the information from the PR (what it fixes, small description)

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/ViewAnnotation.java
```diff
@@ -22,7 +25,13 @@
 
   private static final double MINIMUM_LABEL_WIDTH = 10.0;
   private static final double MINIMUM_LABEL_HEIGHT = 12.0;
+  private static final double MINIMUM_ARROW_WIDTH = 25.0;
```
wouterjsmit, line 28: Minimum arrow with is alrgeer than minimum label width? Sure this is desired? (weird overlapping with many small nodes)

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/ViewAnnotation.java
```diff
@@ -62,9 +71,35 @@ public void addLabel(ArrayList<Node> nodes) {
...
+   * Add an arrow to indicate the annotation direction.
+   *
+   * @param nodes the list of nodes to which to add the arrow.
+   */
+  public void addArrow(ArrayList<Node> nodes) {
+    if (getWidth() < MINIMUM_ARROW_WIDTH || getHeight() < MINIMUM_LABEL_HEIGHT) {
+      return;
+    }
+    ImageView arrow;
+    if (Objects.equals(annotation.strand, "-")) {
```
wouterjsmit, line 83: This should be done with inheritance (`ReverseViewAnnotation extends ViewAnnotation`), Leaving this for now since we want a release.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/ViewAnnotation.java
FarisElghlan, line 28: If the annotation is smaller than `MINIMUM_LABEL_WIDTH` the label (text) isn't drawn. If the annotation is smaller than `MINIMUM_ARROW_WIDTH` the arrow won't be drawn.

PL2/PL2-gui/src/main/java/nl/tudelft/pl2016gr2/gui/view/graph/ViewAnnotation.java
FarisElghlan, line 83: Just to be sure, you do know this is the same as `annotation.strand.equals("-")`, but with a null check right?

wouterjsmit, line 83: Did not explicitly realise, but my point was that deciding to draw either forward or backward arrows should be done with different annotation(view) classes. the null-check is good for safety right now but in general shouldn't be necessary

